PR #73, opened 2020-10-11
Documentation changes

mkg33:
Minor documentation changes: stylistic uniformity, typos, punctuation.

Edit: I've no idea where the added newlines come from (I mean, I haven't added them intentionally). By the way, I'm using Atom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@ julia> Pkg.add("ReservoirComputing")
 ```
 ## Echo State Network example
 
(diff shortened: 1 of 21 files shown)


MartinuzziFrancesco:
The newlines are probably whitespaces that Atom removes automatically, don't see what else could be. Thanks for the corrections on the typos, they were scattered all over

mkg33:
@MartinuzziFrancesco Thanks! I'm glad the Atom thing is not a problem.

PR #76, opened 2020-11-01
Hybrid esn input layer

AnasAbdelR:
This is implementation of the input layer for the hybrid echo state network. It is assumed that the vector of inputs is such that the raw states are followed by the inputs of the model in order. The user specifies what the expected number of inputs are for the model, and some gamma for what percent of  the connections are made between the raw inputs and the reservoir nodes exclusively, while the rest are connected to the reservoir nodes and model inputs. No one reservoir node is connected to more than once. What remains is to allow the prior knowledge to be specified as an argument into an ESN model, and do the same for the output layer as what was done for the input layer

### DIFF
--- a/src/esn_input_layers.jl
+++ b/src/esn_input_layers.jl
@@ -118,3 +118,22 @@ function irrational_sign_input(res_size::Int,
     end
     return W_in
 end
+
+"""
+physics_informed_input(res_size::Int, in_size::Int, sigma::Float64, γ::Float64)
+
+Return a weighted input layer matrix, with random non-zero elements drawn from \$ [-\\text{sigma}, \\text{sigma}] \$, where some γ
+of reservoir nodes are connected exclusively to the raw inputs, and the rest to the outputs of the prior knowledge model , as described in [1].
+
+[1] Jaideep Pathak et al. "Hybrid Forecasting of Chaotic Processes: Using Machine Learning in Conjunction with a Knowledge-Based Model" (2018)
+"""
+function physics_informed_input(res_size::Int,
+        in_size::Int,
+        sigma::Float64,
+        γ::Float64,
+        model_in_size::Int)
+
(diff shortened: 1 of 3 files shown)


src/esn_input_layers.jl
```diff
@@ -118,3 +118,41 @@ function irrational_sign_input(res_size::Int,
     end
     return W_in
 end
+
+"""
+physics_informed_input(res_size::Int, in_size::Int, sigma::Float64, γ::Float64)
+
+Return a weighted input layer matrix, with random non-zero elements drawn from \$ [-\\text{sigma}, \\text{sigma}] \$, where some γ
+of reservoir nodes are connected exclusively to the raw inputs, and the rest to the outputs of the prior knowledge model , as described in [1].
```
ChrisRackauckas, line 126: where is the model specified?

AnasAbdelR: This plan was to have it specified inside the ESN model itself (one of the arguments), so that it is used in both training and prediction as specified in the paper

MartinuzziFrancesco: Yeha, I implemented a couple of ESN variations in the same manner just using a different the layer function in the general ESN constructor. At a quick glance the implementations is good, but I suppose we could wait for a quick comparison with the results of the paper before merging

AnasAbdelR: Implementation

I hope I successfully captured their description. I have attached a screenshot of the part of the paper where they mention the implementation for the input layer. For the entire implementation to be complete still need to:

1. Introduce the prior model into an ESN
2. Do the same for the output weights as what was done in the input
3. make sure the training and prediction work properly

AnasAbdelR: I realised their implementation is to pair the input and knowledge model connections at random with a single reservoir - this isn't exactly the case in the PR - going to update this.

PR #78, opened 2020-11-08
HESN-Constructor

AnasAbdelR:
Here's my implementation of the constructor for the HESN. One of the fields is a prior model function that generates data from a provided model of the form written below as well as the u0, tspan and tsteps to saveat. Wanted to make it so that any kind of differential equation can be accepted and generated given the user configures prior_model_data_generator to his problem:

```julia
function lorenz(du,u,p,t)
    p = [10.0,28.0,8/3]
    du[1] = p[1]*(u[2]-u[1])
    du[2] = u[1]*(p[2]-u[3]) - u[2]
    du[3] = u[1]*u[2] - p[3]*u[3]
end

function prior_model_data_generator(u0, tspan, tsteps, model = lorenz)
    prob = ODEProblem(lorenz, u0, tspan) 
    sol = Array(solve(prob, saveat = tsteps))
    return sol
end

```
The rest of the added fields i.e physics_model_data, u0, tspan, and datasize are used for training and prediction.

For training, my understanding is that the output of the HESN uses a one step ahead prediction value from the prior model relative to the value used for the input, so I made the model solve for one extra time step inside the HESN function:

```julia
    trange = collect(range(tspan[1], tspan[2], length = datasize))
    dt = trange[2]-trange[1]
    tsteps = push!(trange, dt + trange[end])
    tspan_new = (tspan[1], dt+tspan[2])
    physics_model_data = prior_model(u0, tspan_new, tsteps)
    physics_informed_data = vcat(train_data, physics_model_data[:, 1:end-1])
    states = states_matrix(W, W_in, physics_informed_data, alpha, activation, extended_states)
```

This way, when we generate the initial reservoir states, we can use the physics model data in the input at the current time step, and when we train the physics model data we use physics_model_data[:, 2:end] - which is one step ahead. Does this make sense? @ChrisRackauckas @MartinuzziFrancesco 

I have posted my idea of the training function here so you can see how this implementation connects with it

```julia
HESNtrain(ridge::Ridge{T}, hesn::AbstractReservoirComputer; train_data::AbstractArray{Float64} = hesn.train_data) where T<: AbstractFloat = _PIridge(hesn, ridge; train_data = hesn.train_data)

function _PIridge(hesn::AbstractReservoirComputer, ridge::Ridge; train_data::AbstractArray{Float64} = hesn.train_data)

    states_new = nla(hesn.nla_type, hesn.states)
    states_new = vcat(states_new, hesn.physics_model_data[:, 2:end])
    W_out = zeros(Float64, size(train_data, 1), size(states_new, 1))
    for i=1:size(train_data, 1)
        r = RidgeRegression(lambda = ridge.lambda, fit_intercept = false)
        W_out[i,:] = MLJLinearModels.fit(r, states_new', train_data[i,:], solver = ridge.solver)
    end

    return W_out
end
```

Going to pr the training function after the first 2 pieces feel coherent to you guys and make sense (physics input layer and this constructor

### DIFF
--- a/test/constructors/test_hesn_constructors.jl
+++ b/test/constructors/test_hesn_constructors.jl
@@ -39,7 +39,6 @@ function prior_model(u0, tspan, datasize, model = lorenz)
     sol = ones(length(u0), length(tsteps))
     return sol
 end
-concat_train = vcat(train, prior_model(u0, tspan, datasize))
 #constructor 1
 hesn = HESN(W,
     train,
@@ -56,7 +55,7 @@ hesn = HESN(W,
 
 #test constructor
 @test isequal(Integer(floor(approx_res_size/in_size)*in_size), hesn.res_size)
-@test isequal(concat_train, hesn.train_data)
+@test isequal(train, hesn.train_data)
 @test isequal(alpha, hesn.alpha)
 @test isequal(activation, hesn.activation)
 @test isequal(nla_type, hesn.nla_type)
--- a/test/constructors/test_hesn_constructors.jl
+++ b/test/constructors/test_hesn_constructors.jl
@@ -26,7 +26,7 @@ const h_steps = 2
 const train_len = 50
 const predict_len = 12
 data = ones(Float64, in_size-out_size, 100).+γ
-train = data[:, 1:1+train_len-1]
+train = data[:, 1:train_len]
 
 #test = data[:, train_len:train_len+predict_len-1]
 
@@ -60,6 +60,7 @@ hesn = HESN(W,
     nla_type = nla_type,
     extended_states = extended_states)
 
+
 #test constructor
 @test isequal(Integer(floor(approx_res_size/in_size)*in_size), hesn.res_size)
 @test isequal(train, hesn.train_data)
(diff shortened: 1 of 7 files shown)


test/constructors/test_hesn_constructors.jl
```diff
@@ -0,0 +1,78 @@
...
+W_in = ReservoirComputing.physics_informed_input(approx_res_size, in_size, sigma, γ, prior_model_size)
+W = ReservoirComputing.init_reservoir_givendeg(approx_res_size, radius, degree)
+const extended_states = false
+const h_steps = 2
+
+
+const train_len = 50
+const predict_len = 12
+data = ones(Float64, in_size-out_size, 100).+γ
+train = data[:, 1:1+train_len-1]
```
ChrisRackauckas, line 29:
```suggestion
train = data[:, 1:train_len]
```

MartinuzziFrancesco:
Very nice work! At a first glance it does seem that the indexing of physics_model_data[:, 2:end] for the step ahead prediction is correct and the results are also in line with the paper so this is clearly the right direction

Issue #74, opened 2020-10-29
Incorporating prior knowledge into ESNs using hybrid ESN


AnasAbdelR:
Hi,

This paper demonstrates how prior knowledge of the dynamics can be incorporated into the ESN architecture:
https://arxiv.org/pdf/1803.04779.pdf

It had some interesting results produced in the paper below using a reduced order model of a physical process:
https://arxiv.org/pdf/2001.04027.pdf

The basic idea is that instead of having dims = (D_r, M) as an input weight matrix into the reservoir (where D_r is the dimension of the reservoir and M is the dimension of the states) , you have (D_r, M+K) where K is the output dimension of the prior knowledge model. The initialisation of the input weight connections are configured such that some γ of the reservoir nodes are connected to the raw inputs, and the rest are connected to the prior knowledge model. The same is also done for the output read out. 

They use a Erdos-Renyi network for the reservoir, and an additional modification to the reservoir states outputs that empirically helped their situation, but I don't think that will be needed for the general case.

AnasAbdelR:
ESN's with different layers according to the example in the docs looks like this:

```julia
W = DLR(approx_res_size, r)
W_in = init_dense_input_layer(approx_res_size, size(train, 1), sigma)

esndlr = ESN(W,
    train,
    W_in,
    activation = activation,
    alpha = alpha,
    nla_type = nla_type,
    extended_states = extended_states)

W_out = ESNtrain(esndlr, beta)
output = ESNpredict(esndlr, predict_len, W_out)

```
Maintaining the spirit of this workflow, the implementation for this paper should follow a similar style. For the first bit i.e input weights and states of the reservoir, a normal ESN implementation is such that:

![CodeCogsEqn (4)](https://user-images.githubusercontent.com/73660335/98311626-d6dbfd00-1f9d-11eb-931f-bb0e34e89d02.png)

where equation for evolving reservoir states is: r(t+dt) = tanh(W * x(t) + W_in * u(t))
W : dims(r, r) where r is the # of reservoir nodes
x(t) : dims(r, t) where t is the number of time steps
W_in : dims(r, u) where u is the dimensions of the input states
u(t) : dims(u, t)

According to the HESN paper, W_in is such that:

1.  The matrix W_in is to have exactly one nonzero element in each row
2.  The nonzero elements are independently chosen from the uniform distribution on the interval [−σ, σ]
[...]

PR #79, opened 2020-11-08
HESN training implementation

AnasAbdelR:
Here is the implementation of the HESN training that was demonstrated in the  #74 . It only uses ridge regression, might have to implement the other ones was wondering if you guys have any thoughts on how you would approach that given this? Looking forward to your feedback


MartinuzziFrancesco:
Since you have leveraged MLJ for the ridge regression part extending to other linear solvers should be rather straight forward at this point I believe, I'm going to merge your changes and try to extend your train function

Issue #88, opened 2021-03-01
Error when loading in julia 1.6.0-rc1


gbaraldi:
Calling `Using ReservoirComputing` gives the following errors 
```julia
julia> using ReservoirComputing
ERROR: InitError: UndefVarError: Params not defined
Stacktrace:
  [1] top-level scope
    @ none:1
  [2] eval
    @ ./boot.jl:360 [inlined]
  [3] define_typeinf_code2()
    @ IRTools.Inner ~/.julia/packages/IRTools/BpoqK/src/reflection/reflection.jl:21
  [4] __init__()
    @ IRTools.Inner ~/.julia/packages/IRTools/BpoqK/src/IRTools.jl:37
  [5] _include_from_serialized(path::String, depmods::Vector{Any})
    @ Base ./loading.jl:670
  [6] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
    @ Base ./loading.jl:756
  [7] _tryrequire_from_serialized(modkey::Base.PkgId, build_id::UInt64, modpath::String)
    @ Base ./loading.jl:685
  [8] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
    @ Base ./loading.jl:745
  [9] _tryrequire_from_serialized(modkey::Base.PkgId, build_id::UInt64, modpath::String)
    @ Base ./loading.jl:685
 [10] _require_search_from_serializedjulia> using ReservoirComputing
ERROR: InitError: UndefVarError: Params not defined
Stacktrace:
  [1] top-level scope
    @ none:1
  [2] eval
    @ ./boot.jl:360 [inlined]
  [3] define_typeinf_code2()
    @ IRTools.Inner ~/.julia/packages/IRTools/BpoqK/src/reflection/reflection.jl:21
  [4] __init__()
    @ IRTools.Inner ~/.julia/packages/IRTools/BpoqK/src/IRTools.jl:37
  [5] _include_from_serialized(path::String, depmods::Vector{Any})
    @ Base ./loading.jl:670
  [6] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
    @ Base ./loading.jl:756
  [7] _tryrequire_from_serialized(modkey::Base.PkgId, build_id::UInt64, modpath::String)
    @ Base ./loading.jl:685
  [8] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
[...]

PR #96, opened 2021-06-11
Fixes for LoadError

MartinuzziFrancesco:


MartinuzziFrancesco:
This solves #88.

ChrisRackauckas:
That's not a fix. But I guess your plan is to just trigger compathelper again?

MartinuzziFrancesco:
yeha that's the idea for the moment

Issue #124, opened 2022-06-13
Tests are failing for Julia 1.7


MartinuzziFrancesco:
Between 1.6 and 1.7 something changed on random numbers generation. The tests at the moment are run using the lts version (1.6) but failing on 1.7. 

Ideally the tests should be robust enough to be run on both versions

Issue #120, opened 2022-05-04
Change `nla()` to reduce allocations


MartinuzziFrancesco:
At the moment the `nla()` function takes an array as input and returns the modified array as output, but this creates a lot of allocations inside `for` loops. The new function call `nla!()` should take in the array to be modified and a preallocated array to store the modified array in. This should remove unneeded allocations and speed up computations.

PR #125, opened 2022-06-20
Fix nla allocation situation

Chronum94:
This is an attempt at fixing #120. Preemptive apologies if the code isn't up to best practices, I just saw the issue lying around and tried to hack together a solution.

I'm on Julia 1.7.3, so there are some tests failing ( #124 ), but the _same_ tests fail before and after, so at least we have that going.

The speedups are most noticeable for the case of many calls with small arrays, with ~1.2-1.4x speedup for 1k x 1k arrays. I don't know what the arrays are supposed to look like, so I just tested out a bunch of sizes.

I'm not really a common Julia package contributor, so I can attempt edits if there's something I've done horrendously wrong in terms of style/best practices.

### DIFF
--- a/src/states.jl
+++ b/src/states.jl
@@ -93,6 +93,12 @@ function nla(::NLADefault, x)
     return x
 end
 
+function nla(nla_type, x_old)
+    x_new = similar(x_old)
+    nla!(nla_type, x_old, x_new)
+    return x_new
+end
+
 """
     NLAT1()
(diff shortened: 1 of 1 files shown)


MartinuzziFrancesco:
thanks for taking a stab at this, at a first glance it seems to be on point! the `nla` functions are explicitly tested so as long as those tests pass we should be good to go

Issue #127, opened 2022-07-07
Alternatives for GaussianProcesses.jl


MartinuzziFrancesco:
For the Gaussian regression at the moment the package uses [GaussianProcesses.jl](https://github.com/STOR-i/GaussianProcesses.jl) and it seems to be the blocker for a bunch of dependencies. Ideally it should be dropped in favor of AbstractGPs.jl

Related to issues https://github.com/SciML/SciMLDocs/pull/23 and https://github.com/SciML/SciMLDocs/issues/2

PR #129, opened 2022-08-11
Removing GaussianProcesses

MartinuzziFrancesco:
Removing GaussianProcesses as a first step to deal with #127


MartinuzziFrancesco:
@ChrisRackauckas I'd merge this and tag a breaking change to deal with the getting stuck on 0.5 problem

PR #143, opened 2022-12-10
Docs work

MartinuzziFrancesco:
Adding `@example` and fixing minor things


ChrisRackauckas:
Skip the GPU one for now. That'll need an issue and a change to the buildkite

MartinuzziFrancesco:
ok thanks I didn't know how to proceed on that one

ChrisRackauckas:
Is there a good tutorial to elevate to a `getting_started`?

ChrisRackauckas:
Also, please split tutorials and examples

MartinuzziFrancesco:
I think the Lorenz example should work well as a `getting_started`. As a rule of thumb what would be a tutorial and what an example? at the moment the structure uses the examples as tutorials

ChrisRackauckas:
https://github.com/SciML/ModelingToolkit.jl/pull/1993

Tutorials should be around specific topics of the package, demonstrating something "most" people should know. Examples are more centered around showing a complex example, maybe with less text. If you have two tutorials around the same topic, one should be kept as tutorial and the other should be demoted to example.

Lorenz is a good getting started.

MartinuzziFrancesco:
Ok I see, that'll take some reorganization. I think that we could merge this pr when everything is up to speed and I'll open a new one for the restructure of the docs specifically.

I can't figure out where [this](https://github.com/SciML/ReservoirComputing.jl/actions/runs/3669953922/jobs/6204152002#step:5:29) is coming from

Issue #202, opened 2024-02-24
Feed data to `train` instead of `ESN`


MartinuzziFrancesco:
Pretty much title, the goal is to divide the definition of the model and the actual training. We could also add an `instantiate` function to run the data through the ESN to obtain the states without the training

Issue #194, opened 2024-01-16
Add external dependencies for training strategies


MartinuzziFrancesco:
Packages like LIBSVM or MLJLinearModels can be external (weak) dependencies.

Add a lower level function that takes the place of `_train` structured like `train(algo::TrainingAlgo, states, target; kwargs...)`. The upper level function still maintains `train(esn::AbstractReservoirComputer, target_data, training_algo; kwargs...)`. Additionally include a version that has `train(esn::AbstractReservoirComputer, training_data, target_data, training_algo; kwargs...)` to prepare for #202

PR #205, opened 2024-02-28
Changes to training

MartinuzziFrancesco:
Addressing #194 and hopefully #202


MartinuzziFrancesco:
We need more drastic changes for #202, I'll merge this as is and I'll address #202 in a future pr

PR #213, opened 2024-03-11
Move initializers to WeightInitializers.jl

### LINKED ISSUE #212: Move initializers to WeightInitializers
Following issue [21](https://github.com/LuxDL/WeightInitializers.jl/issues/21) in WeightInitializers.jl we should do something similar for the initializers that we have here

MartinuzziFrancesco:
This PR will need the [PR](https://github.com/LuxDL/WeightInitializers.jl/pull/23) on WeightInitializers.jl to be merged first


MartinuzziFrancesco:
Closing this as the current initializer set is highly field-specific to reservoir computing (delay lines, cycle reservoirs, chaotic-map/chebyshev/logistic encodings, LSM connectivity). Moving all of that into a generic, framework-agnostic package like WeightInitializers.jl wouldn't provide any benefit I can immediately see. 

The initializers in ReservoirComputing.jl will continue to use WeightInitializers.jl as guidelines for implementation and testing.

PR #267, opened 2025-03-14
Generalizing minimal initializers

### LINKED ISSUE #265: Reduce code reuse in inits
The deterministic initializers have a lot of structures in common. I think we can reduce code reuse by building functions such as

```julia
function delay_line!(reservoir_matrix::AbstractMatrix, weight::Number)
    for idx in first(axes(reservoir_matrix, 1)):(last(axes(reservoir_matrix, 1)) - 1)
        reservoir_matrix[idx + 1, idx] = weight
    end
    return reservoir_matrix
end
```

This way we can even have more control over the weights by having a dispatch over a vector of values

```julia
function delay_line!(reservoir_matrix::AbstractMatrix, weight::AbstractVector)
    for idx in first(axes(reservoir_matrix, 1)):(last(axes(reservoir_matrix, 1)) - 1)
        reservoir_matrix[idx + 1, idx] = weight[idx]
    end
    return reservoir_matrix
end
```

MartinuzziFrancesco:
## Checklist

## Additional context

Closes #265. I am thinking if the building pieces need to be user facing too. I need to update existing documentation to reflect the new flexibility of the initializers (option of using vector of values for `weight`). Add docstrings to the building pieces too if they become public api.


MartinuzziFrancesco:
I think that we can generalize the minimal reservoir inits even more now by also having some random sign changes there. For this we need to format `irrational` and `bernoulli` to simply modify the arrays/matrices in place. Something like

```julia
function bernoulli!(rng::AbstractRNG, vecormat::AbstractVecOrMat; p::Number=0.5)
    for idx in eachindex(vecormat)
        if rand(rng) > p # keeping current approach consistent 
            vecormat[idx] = - vecormat[idx]
        end
    end
end
```

and then having the option to change signs in the inits functions like so

```julia
function delay_line!(rng::AbstractRNG, reservoir_matrix::AbstractMatrix, weight::AbstractVector,
        shift::Int; sampling_type=:default, kwargs...) # :bernoulli, :irrational
    for idx in first(axes(reservoir_matrix, 1)):(last(axes(reservoir_matrix, 1)) - shift)
        reservoir_matrix[idx + shift, idx] = weight[idx]
    end
    f_sample = getfield(@__MODULE__, sampling_type)
    f_sample!(rng, reservoir_matrix; kwargs...)
end
```

MartinuzziFrancesco:
Not really sure why docs are failing, locally they compile just fine. @ChrisRackauckas I can't access the buildkite log it seems, is there any way to get access?

ChrisRackauckas:
Pipeline is public now.

MartinuzziFrancesco:
Awesome, thanks

MartinuzziFrancesco:
all seems to be fine now

Issue #353, opened 2026-01-05
Neuromorphic reservoir computing


MartinuzziFrancesco:
## Paper

https://doi.org/10.1063/5.0282708

## Equations

Model 1:

```math
\begin{aligned}
\mathbf{x}(t)
&= b_{\mathrm{ex}}\, \phi\!\left(
\mathbf{W}_{\text{in}}\, \mathbf{u}(t)
+ a_{\mathrm{ex}}\, \mathbf{A}\, \mathbf{x}(t-1)
\right)
- b_{\mathrm{inh}}\, \phi\!\left(
\mathbf{W}_{\text{in}}\, \mathbf{u}(t)
+ a_{\mathrm{inh}}\, \mathbf{A}\, \mathbf{x}(t-1)
\right).
\end{aligned}
```

Model 2:

```math
\begin{aligned}
\mathbf{x}(t)
&= b_{\mathrm{ex}}\, \phi\!\left(a_{\mathrm{ex}}\, \mathbf{A}\, \mathbf{x}(t-1)\right)
- b_{\mathrm{inh}}\, \phi\!\left(a_{\mathrm{inh}}\, \mathbf{A}\, \mathbf{x}(t-1)\right)
+ g\, \mathbf{W}_{\text{in}}\, \mathbf{u}(t).
\end{aligned}
```

## Official implementation

none

PR #366, opened 2026-01-18
Feat: Add EIESNAdditive (Model 2) with tests and docs

MarcoCerino23:
This PR implements the **EIESN Model 2 (Additive Input)** variant as discussed in Issue #353.

In this model, the input is added linearly to the reservoir state **after** the non-linear activation, rather than being passed through the activation function along with the recurrent part.

### Changes:
- Added `EIESNAdditiveCell` struct in `src/layers/eiesn_additive_cell.jl`.
- Added `EIESNAdditive` model wrapper in `src/models/eiesn_additive.jl`.
- Implemented the specific math: $x(t) = \dots + g W_{in} u(t)$.
- Added `input_scale` parameter (representing $g$).
- Added comprehensive tests for both the cell and the model.
- Updated documentation in `layers.md` and `models.md`.
- Exported new types in `ReservoirComputing.jl`.

## Checklist

## Additional context
Ref: #353.
This complements the previously merged Model 1 implementation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ReservoirComputing"
 uuid = "7c2d2b1e-3dd4-11ea-355a-8f6a8116e294"
-version = "0.12.12"
+version = "0.12.13"
 authors = ["Francesco Martinuzzi"]
 
 [deps]
--- a/src/layers/additive_eiesn_cell.jl
+++ b/src/layers/additive_eiesn_cell.jl
@@ -1,12 +1,15 @@
 @doc raw"""
     AdditiveEIESNCell(in_dims => out_dims, [activation]; kwargs...)
 
-Excitatory-Inhibitory Echo State Network (EIESN) cell with additive input [Panahi2025](@cite).
+Excitatory-Inhibitory Echo State Network (EIESN) cell with additive input
+[Panahi2025](@cite).
 
 ```math
 \mathbf{x}(t) =
-b_{\mathrm{ex}} \, \phi_{\mathrm{ex}}\!\left(a_{\mathrm{ex}} \mathbf{A} \mathbf{x}(t-1) + \mathbf{\beta}_{\mathrm{ex}}\right)
-- b_{\mathrm{inh}} \, \phi_{\mathrm{inh}}\!\left(a_{\mathrm{inh}} \mathbf{A} \mathbf{x}(t-1) + \mathbf{\beta}_{\mathrm{inh}}\right)
+b_{\mathrm{ex}} \, \phi_{\mathrm{ex}}\!\left(
+  a_{\mathrm{ex}} \mathbf{A} \mathbf{x}(t-1) + \mathbf{\beta}_{\mathrm{ex}}\right)
+- b_{\mathrm{inh}} \, \phi_{\mathrm{inh}}\!\left(
+  a_{\mathrm{inh}} \mathbf{A} \mathbf{x}(t-1) + \mathbf{\beta}_{\mathrm{inh}}\right)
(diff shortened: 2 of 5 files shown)


src/layers/eiesn_additive_cell.jl
```diff
@@ -0,0 +1,133 @@
...
+
+## States
+
+Created by `initialstates(rng, cell)`:
+
+    - `rng`: Replicated RNG used to initialize the hidden state when an external
+        state is not provided.
+"""
+
+@concrete struct EIESNAdditiveCell <: AbstractReservoirRecurrentCell
```
MartinuzziFrancesco:
```suggestion
@concrete struct EIESNAdditiveCell <: AbstractEchoStateNetworkCell
```

src/layers/eiesn_additive_cell.jl
```diff
@@ -0,0 +1,133 @@
...
+    - `reservoir_matrix :: (out_dims × out_dims)` — $\mathbf{A}$.
+
+## States
+
+Created by `initialstates(rng, cell)`:
+
+    - `rng`: Replicated RNG used to initialize the hidden state when an external
+        state is not provided.
+"""
+
```
MartinuzziFrancesco: delete the space to make docs build

MartinuzziFrancesco:
Nice, feels like it's almost ready! maybe change the name to `AdditiveEIESN` and `AdditiveEIESNCell` to maintain the style `ModificationModel` that we have going in the library. Also, I don't see a bias. do they really not use one in the paper?

MarcoCerino23:
Done.
Naming: I've renamed all structs and files to AdditiveEIESN and AdditiveEIESNCell to match the library style.
Bias: I double-checked the reference paper (Panahi et al., 2025). Specifically, Equation (5) in the preprint explicitly defines the update rule as a linear addition of the input term without any bias vector.

MarcoCerino23:
Thanks for the feedback. Regarding the recurrent part, I noticed that the EIESN formulation requires scaling the matrix product before adding the bias: $\alpha (\mathbf{A}\mathbf{x}) + \mathbf{b}$.
ESNCell uses dense_bias because it computes $\mathbf{W}\mathbf{x} + \mathbf{b}$. To achieve the equivalent result here using dense_bias, I would need to scale the matrix itself beforehand (e.g., passing α .* A as the first argument).
I suspect this might be less efficient (especially with sparse matrices) compared to the current implementation where I scale the resulting vector. Do you agree with keeping the explicit calculation for efficiency, or would you prefer using dense_bias with matrix scaling to clean up the code?

MarcoCerino23:
That's a great observation regarding associativity, and it would work ideally for a standard ESN.
However, there is a structural constraint in the EIESN architecture that prevents this optimization:
1. **Dual Biases**: Unlike a standard ESN, we have two distinct biases (bias_ex and bias_inh) for the excitatory and inhibitory branches.
2. **Shared Input**: We calculate the input projection W_in * u once and share it across both branches for efficiency.
If I were to use dense_bias to merge bias_ex into the input, the inhibitory branch would receive the wrong bias (or bias_ex + bias_inh). I would effectively have to compute the input projection **twice** (once with bias_ex, once with bias_inh) using dense_bias, which doubles the computational cost of the input layer.
For AdditiveEIESNCell(Model 2), the input is even added **outside** the non-linearity, so merging the recurrent bias into the input matrix is mathematically impossible there.

MartinuzziFrancesco:
I see, that blocks a lot of potential solutions. Then I think your implementation is the right approach, there's probably less overhead for the ifs that there would be to compute the same matmul twice. Just up the version, and make sure that the docstrings also do not go over 92 cols and we can merge this!

Issue #284, opened 2025-03-31
Wigner reservoir


MartinuzziFrancesco:
## Description

The diagonal elements are distributed as 

```math
$$
W_{\text{w},ii} \sim \mathcal{N}(0, \rho_1^2 / n),
$$ 
```
whereas off-diagonal elements are created as:
```math
$$
W_{\text{w},ij} = W_{\text{r},ji} \sim \mathcal{N} \left( 0, \frac{\rho_2^2}{n} \right), \quad i \ne j. \tag{10}
$$
```

## Source

I found this initialization in [Verzelli2022](https://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=9366982&casa_token=glzRvGJcMlwAAAAA:XsQVxU-TOv_GnkqweRWHxjS26NevgOaz_I_XEKNH0uDriDm1IUHfwlpAd2xrzb2HoLNHGPH8dunI&tag=1) which mentions [Inubushi2017](https://www.nature.com/articles/s41598-017-10257-6.pdf) as the original source.

MartinuzziFrancesco:
closed by #413

PR #412, opened 2026-03-25
Adding a Wigner matrix initialiser for #284

am-1118:
This PR implements the Wigner matrix initialisation to resolve Issue #284.

##  Changes Made
* Added `wigner_init` to the matrix initialisers.
* Utilised `LinearAlgebra.Symmetric` to make the matrix symmetric.
* Integrated the standard `check_res_size`, `scale_radius!`, and `check_inf_nan` utilities.
* Added a `@testset` block to verify reservoir dimensions, data type, and reservoir symmetry.

### DIFF
--- a/src/inits/inits_reservoir.jl
+++ b/src/inits/inits_reservoir.jl
@@ -2538,38 +2538,51 @@ end
 
 function wigner_init(
         rng::AbstractRNG, ::Type{T}, dims::Integer...;
-        radius::Number = T(1.0), std::Number = T(1.0)
+        radius::Number = T(1.0), diagonal_std::Number, off_diagonal_std::Number,
+        return_symmetric::Bool = true
     ) where {T <: Number}
     # 1. Dimension check : Reservoir has to be a square matrix
     check_res_size(dims...)
     res_size = dims[1]  
 
-    # 2. Initialise the empty matrix using the requested type T
-    W = zeros(T, res_size, res_size)
+    # 2. Initialise the zeros matrix with type T
+    reservoir_matrix = zeros(T, res_size, res_size)
 
     # 3. Populating the matrix
-    # Diagonal elements sampled from N(0, 2*std^2)
-    # Off-diagonal elements sampled from N(0, std^2)
-    for i in 1 : res_size
-        for j in 1 : res_size
-            if i==j
-                # Diagonal element
-                W[i, j] = randn(rng, T) * T(sqrt(2)) * T(std)
-            else
-                # Off-diagonal elements (upper triangular part)
-                W[i, j] = randn(rng, T) * T(std)
+    # Diagonal elements sampled from N(0, diagonal_std^2)
+    # Off-diagonal elements sampled from N(0, off_diagonal_std^2)
+    
+    # Default symmetric case
+    if return_symmetric
+        # Efficient upper-triangle fill for symmetric matrices
+        for i in 1:res_size
+            for j in i:res_size
+                if i == j
+                    reservoir_matrix[i, j] = DeviceAgnostic.randn(rng, T) * T(diag_std)
+                else
+                    reservoir_matrix[i, j] = DeviceAgnostic.randn(rng, T) * T(off_diag_std)
+                end
+            end
+        end
+        # Mirror the matrix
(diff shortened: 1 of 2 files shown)


src/inits/inits_reservoir.jl
```diff
@@ -2533,3 +2533,43 @@ for initializer in (
...
+    W = zeros(T, res_size, res_size)
+
+    # 3. Populating the matrix
+    # Diagonal elements sampled from N(0, 2*std^2)
+    # Off-diagonal elements sampled from N(0, std^2)
+    for i in 1 : res_size
+        for j in 1 : res_size
+            if i==j
+                # Diagonal element
+                W[i, j] = randn(rng, T) * T(sqrt(2)) * T(std)
```
MartinuzziFrancesco: `randn(rng, T)` sohuld be taken from `DeviceAgnostic`

also I don't understand why `T(sqrt(2))`

src/inits/inits_reservoir.jl
```diff
@@ -2533,3 +2533,43 @@ for initializer in (
...
+            else
+                # Off-diagonal elements (upper triangular part)
+                W[i, j] = randn(rng, T) * T(std)
+            end
+        end
+    end
+
+    # Make the matrix symmetric
+    W = Symmetric(W)
```
MartinuzziFrancesco: add this as a kwarg option that the user can choose `return_symmetric::Bool = false`

src/inits/inits_reservoir.jl
```diff
@@ -2533,3 +2533,43 @@ for initializer in (
...
+
+    # 4. Scaling the spectral radius to the user-specified value
+    W = scale_radius!(W, T(radius))
+
+    # 5. Check for NaN or Inf values
+    check_inf_nan(W)
```
MartinuzziFrancesco: I don't think this is necessary, this issue usually arises from the sparsity

src/inits/inits_reservoir.jl
```diff
@@ -2533,3 +2533,43 @@ for initializer in (
...
+function wigner_init(
+        rng::AbstractRNG, ::Type{T}, dims::Integer...;
+        radius::Number = T(1.0), std::Number = T(1.0)
+    ) where {T <: Number}
+    # 1. Dimension check : Reservoir has to be a square matrix
+    check_res_size(dims...)
+    res_size = dims[1]  
+
+    # 2. Initialise the empty matrix using the requested type T
+    W = zeros(T, res_size, res_size)
```
MartinuzziFrancesco: try to avoid single letter naming if possible W -> reservoir_matrix. follow the conventions of the other initializers so that it makes it easier to search for stuff in the file

src/inits/inits_reservoir.jl
MartinuzziFrancesco: also need to import this from LinearAlgebra [here](https://github.com/SciML/ReservoirComputing.jl/blob/master/src/ReservoirComputing.jl#L5)

MartinuzziFrancesco (changes requested):

src/inits/inits_reservoir.jl
am-1118: For the standard Wigner matrix, T(sqrt(2)) is for the diagonal elements as they are sampled from N(0, 2*variance), which implies sqrt(2) * std.  I will modify the arguments to accept two arbitrary std values for the diagonal and off-diagonal distributions, as described in the issue. Modified rand -> DeviceAgnostic.rand

src/inits/inits_reservoir.jl
am-1118: Done

src/inits/inits_reservoir.jl
am-1118: Should I remove that line ? Or add some other check?

src/inits/inits_reservoir.jl
am-1118: I've added the return_symmetric argument, but I've kept the default case as true, which the user can set to false.
using LinearAlgebra: eigvals, I, qr, Diagonal, diag, mul!
Should I paste the above line into the inits_reservoir.jl code?

MartinuzziFrancesco:
closed in favor of #413

PR #414, opened 2026-03-28
Add ReservoirComputingBenchmarks library

Saswatsusmoy:
## Summary

Initial implementation of the `ReservoirComputingBenchmarks` library in `lib/` as discussed in #398. Implements the first three core benchmark tasks:

- **Memory Capacity** (Jaeger 2002) — linear memory evaluation with per-delay and total MC
- **NARMA** (Atiya & Parlos 2000) — target generation (orders 2, 10, 30) and evaluation with NMSE/RNMSE/MSE
- **Information Processing Capacity** (Dambre et al. 2012) — full capacity decomposition via Legendre polynomial basis functions

### Design decisions

- **Model-agnostic API**: all functions take raw `(input, states)` arrays, no RC.jl dependency
- **Lives in `lib/`**: separate package under the monorepo as discussed
- **Zero external dependencies**: only `LinearAlgebra` + `Statistics` (stdlib)
- **Cholesky-based ridge regression** with factorization reuse — the Gram matrix is factorized once and reused across all delays/basis functions (MC: ~6ms, IPC: ~6ms for 108 basis functions)
- **No plotting** — computation only, as discussed
- Preset NARMA coefficients for orders 2, 10, 30; custom coefficients supported for arbitrary orders
- Input validation with clear error messages; warnings for degenerate cases (NaN, divergence)

### Exported API

| Function | Signature | Returns |
|----------|-----------|---------|
| `memory_capacity` | `(input, states; max_delay, train_ratio, reg)` | `(total, delays)` |
| `generate_narma` | `(input; order, alpha, beta, gamma, delta, normalize)` | target vector |
| `narma` | `(input, states; order, metric, train_ratio, reg, washout)` | `(score, metric, target)` |
| `ipc` | `(input, states; max_delay, max_degree, cross_terms, reg)` | `(total, linear, nonlinear, by_degree, by_delay, basis_capacities, theoretical_max)` |

Closes #398 (partially — remaining 5 tasks can be added incrementally).

## Test plan

- [x] 85 unit tests passing (`Pkg.test()`)
- [x] End-to-end integration test with simulated 100-node ESN
- [x] Edge cases: small reservoirs, short signals, variable train ratios, variable regularization
- [x] Validation: 12 error cases (mismatched dims, invalid params, out-of-range values)
- [x] Reproducibility: deterministic results on repeated calls
- [x] NARMA divergence detection with warnings

### DIFF
--- a/lib/ReservoirComputingBenchmarks/src/ipc.jl
+++ b/lib/ReservoirComputingBenchmarks/src/ipc.jl
@@ -13,7 +13,7 @@ z_l(t) = \prod_i \tilde{P}_{d_i}\!\bigl(u(t - \tau_i)\bigr)
 
 where ``\tilde{P}_n(x) = P_n(x)\sqrt{(2n+1)/2}`` is the normalized Legendre
 polynomial of degree ``n``.  For each basis function, a ridge regression readout
-is trained, and the capacity is the squared correlation between the target and
+is trained, and the capacity is the squared correlation between target and
 prediction.
 
 The total capacity satisfies ``C_{\text{total}} \leq N`` where ``N`` is the
@@ -76,11 +76,11 @@ function ipc(
     mtd = something(max_total_degree, max_degree)
 
     # Precompute normalized Legendre polynomial values for all (degree, delay)
-    poly_cache = Dict{Tuple{Int, Int}, Vector{Float64}}()
+    poly_cache = Matrix{Vector{Float64}}(undef, max_delay, max_degree)
     for delay in 1:max_delay
         for degree in 1:max_degree
             delayed = @view input[(max_delay + 1 - delay):(T - delay)]
-            poly_cache[(delay, degree)] = _normalized_legendre.(degree, delayed)
+            poly_cache[delay, degree] = map(x -> _normalized_legendre(degree, x), delayed)
         end
     end
 
@@ -90,43 +90,43 @@ function ipc(
     # Valid time range (after max_delay to avoid edge effects)
     valid = (max_delay + 1):T
     T_valid = length(valid)
-    X = collect(states[:, valid]')  # (T_valid, n_features)
+    X = Matrix{Float64}(undef, T_valid, size(states, 1))
+    copyto!(X, view(states, :, valid)')
 
     train_idx, test_idx = _train_test_split(T_valid, train_ratio)
-    X_train = X[train_idx, :]
-    X_test = X[test_idx, :]
+    X_train = view(X, train_idx, :)
+    X_test = view(X, test_idx, :)
+    n_test = length(test_idx)
 
-    # Pre-compute Cholesky factorization — reused across all basis functions
     rf = _ridge_factor(X_train; reg = reg)
 
     results = Vector{NamedTuple{(:terms, :degree, :capacity), Tuple{Vector{Tuple{Int, Int}}, Int, Float64}}}()
+    sizehint!(results, length(basis_functions))
     by_degree = Dict{Int, Float64}()
     by_delay = Dict{Int, Float64}()
 
-    # Pre-allocate target buffer — reused via fill!
     target = Vector{Float64}(undef, T_valid)
+    y_pred = Vector{Float64}(undef, n_test)
 
-    for terms in basis_functions
+    @inbounds for terms in basis_functions
         total_degree = sum(d for (_, d) in terms)
 
-        # Compute target: product of normalized Legendre polynomials
         fill!(target, 1.0)
         for (delay, degree) in terms
-            target .*= poly_cache[(delay, degree)]
+            target .*= poly_cache[delay, degree]
         end
 
-        y_train = target[train_idx]
-        y_test = target[test_idx]
+        y_train = view(target, train_idx)
+        y_test = view(target, test_idx)
 
-        w = _ridge_solve(rf, X_train, y_train)
-        y_pred = X_test * w
+        w = _ridge_solve!(rf, X_train, y_train)
+        mul!(y_pred, X_test, w)
 
         cap = _squared_correlation(y_test, y_pred)
 
         push!(results, (terms = terms, degree = total_degree, capacity = cap))
         by_degree[total_degree] = get(by_degree, total_degree, 0.0) + cap
 
-        # Track per-delay capacity for single-variable terms
         if length(terms) == 1
             delay = terms[1][1]
             by_delay[delay] = get(by_delay, delay, 0.0) + cap
@@ -153,14 +153,14 @@ end
 """
     _legendre(n, x)
 
-Evaluate the Legendre polynomial ``P_n(x)`` using the three-term recurrence.
+Evaluate Legendre polynomial ``P_n(x)`` using three-term recurrence.
 """
-function _legendre(n::Int, x::Real)
+@inline function _legendre(n::Int, x::Real)
     @assert n >= 0 "Legendre polynomial degree must be non-negative, got $n"
     n == 0 && return one(float(x))
     n == 1 && return float(x)
     p_prev, p_curr = one(float(x)), float(x)
-    for k in 1:(n - 1)
+    @inbounds for k in 1:(n - 1)
         p_next = ((2k + 1) * x * p_curr - k * p_prev) / (k + 1)
         p_prev, p_curr = p_curr, p_next
     end
@@ -173,7 +173,7 @@ end
 Evaluate the orthonormalized Legendre polynomial
 ``\\tilde{P}_n(x) = P_n(x) \\sqrt{(2n+1)/2}`` on ``[-1,1]``.
 """
-function _normalized_legendre(n::Int, x::Real)
+@inline function _normalized_legendre(n::Int, x::Real)
     return _legendre(n, x) * sqrt((2n + 1) / 2)
 end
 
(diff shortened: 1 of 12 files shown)


MartinuzziFrancesco:
I have went through the code and I'll provide some detailed comments tomorrow. The implementations look correct, but there are a couple of details were the code could be made more flexible. 

In the meantime I am thinking that you probably were right in the issue and the benchmarks could be more linked to the models, since that would reduce the code reuse. Without rewriting the functions for the moment, what do you think the signatures of the functions would be like?

Saswatsusmoy:
For the model-integrated signatures, I was thinking we keep the existing raw-array methods as the base and add dispatch methods that accept RC.jl models directly. 

Something like:                           
   
``` julia                                                                                                                                                                                                   
# existing (stays as-is, model-agnostic)                                                                                                                                                                   
   memory_capacity(input::AbstractVector, states::AbstractMatrix; max_delay=30, reg=1.0)                                                                                                                        
                                                                                                                                                                                                               
# new dispatch — handles input generation + state collection internally                                                                                                                                      
   memory_capacity(model::AbstractReservoirComputer, ps, st;                                                                                                                                                    
       T=3000, rng=Random.default_rng(), max_delay=30, reg=1.0)
```                                                                                                                                                
                                                                                                                                                                                                             
The model method would generate the input signal, call collectstates(model, reshape(input, 1, :), ps, st) to get the states, then delegate to the raw-array method. Same pattern for narma and ipc:          
   
 ```julia
narma(model::AbstractReservoirComputer, ps, st;                                                                                                                                                              
    T=3000, rng=Random.default_rng(), order=10, metric=:nmse, reg=1.0)                                                                                                                                                                                       

ipc(model::AbstractReservoirComputer, ps, st;
    T=3000, rng=Random.default_rng(), max_delay=10, max_degree=3, reg=1.0)
```                                                                                                                               
                                                                                                                                                                                                             
So the user workflow becomes just                                                                                                                                              
   
```julia
model = ESN(1, 100, 1)                                                                                                                                                                                       
ps, st = setup(rng, model)                                                                                                                                                                                 mc = memory_capacity(model, ps, st; T=3000, max_delay=50)     
```                                                                                                                                               
   
For the dependency question I think this could work nicely as a package extension that activates when ReservoirComputing is loaded, so the base package stays zero-dependency and the RC.jl integration is opt-in. 
  

What do you think?

MartinuzziFrancesco:
yeha I think that that would be a nice way to approach it. I will go over the code and provide you with some comments

Saswatsusmoy:
Any updates on this?

lib/ReservoirComputingBenchmarks/src/narma.jl
```diff
@@ -0,0 +1,190 @@
...
+    score = if metric === :nmse
+        _nmse(y_test, y_pred)
+    elseif metric === :rnmse
+        nmse_val = _nmse(y_test, y_pred)
+        nmse_val < 0 ? 0.0 : sqrt(nmse_val)
+    elseif metric === :mse
+        mean((y_test .- y_pred) .^ 2)
+    else
+        throw(ArgumentError("Unknown metric :$metric. Use :nmse, :rnmse, or :mse."))
+    end
```
MartinuzziFrancesco: you can avoid this by letting `metric` accept any function with a `(test, pred)` signature. This way the function can also support any metric the user wants to use

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
I was offline the past few days, I'm testing the code at the moment

Saswatsusmoy:
Hi @MartinuzziFrancesco

Pushed the rest of issue #398 in 3b84876c. All eight benchmarks now have an array form and a model dispatch through a package extension that activates when ReservoirComputing is loaded. The base library stays zero dep so anyone can use it without pulling in RC.jl.

The extension adds (model, ps, st) dispatch on memory_capacity, nonlinear_memory, nonlinear_transformation, sin_approximation, narma, ipc, kernel_rank and generalization_rank. It generates a uniform [-1, 1] driving signal of length T, collects states via collectstates, and forwards to the array form. There is also a friendly in_dims == 1 check so users get a clear error instead of an opaque shape mismatch.

Test count is 180 covering happy path, validation errors, reproducibility, a deterministic generalization rank invariant (perturbation 0 gives rank 1) and cross model dispatch on ES2N and EuSN.

A few things I left out so this PR stays focused. Flagging them as follow ups so we can plan them separately.

1. The lib subpackage has no CI workflow yet so PRs touching it can break it without notice. Want me to add one in a follow up
2. No user facing docs in docs/src for the benchmarks. Happy to draft a tutorial page and an API page
3. No Aqua.jl or JET.jl quality checks. SciML standard for new packages so probably worth adding before release
4. Float32 and GPU paths are untested. Tests run CPU Float64 only
5. Cross model tests cover ESN ES2N and EuSN. The dispatch should work for NGRC RECA SVESM and LIFESN through the abstract type but nothing confirms it
6. Registration in the General registry whenever you want to cut v0.1.0

Let me know which of these you want me to take on next and which you want to defer.

lib/ReservoirComputingBenchmarks/src/ipc.jl
```diff
@@ -0,0 +1,215 @@
...
+
+    rf = _ridge_factor(X_train; reg = reg)
+
+    results = Vector{NamedTuple{(:terms, :degree, :capacity), Tuple{Vector{Tuple{Int, Int}}, Int, Float64}}}()
+    sizehint!(results, length(basis_functions))
+    by_degree = Dict{Int, Float64}()
+    by_delay = Dict{Int, Float64}()
+
+    target = Vector{Float64}(undef, T_valid)
+    y_pred = Vector{Float64}(undef, n_test)
```
MartinuzziFrancesco: don't force Float64 if you can infer it from input/states

lib/ReservoirComputingBenchmarks/src/ipc.jl
```diff
@@ -0,0 +1,215 @@
...
+    X_train = view(X, train_idx, :)
+    X_test = view(X, test_idx, :)
+    n_test = length(test_idx)
+
+    rf = _ridge_factor(X_train; reg = reg)
+
+    results = Vector{NamedTuple{(:terms, :degree, :capacity), Tuple{Vector{Tuple{Int, Int}}, Int, Float64}}}()
+    sizehint!(results, length(basis_functions))
+    by_degree = Dict{Int, Float64}()
+    by_delay = Dict{Int, Float64}()
```
MartinuzziFrancesco: don't force Float64 if you can infer it from input/states

lib/ReservoirComputingBenchmarks/src/ipc.jl
```diff
@@ -0,0 +1,215 @@
...
+        end
+    end
+
+    # Generate basis functions
+    basis_functions = _enumerate_basis_functions(max_delay, max_degree, mtd, cross_terms)
+
+    # Valid time range (after max_delay to avoid edge effects)
+    valid = (max_delay + 1):T
+    T_valid = length(valid)
+    X = Matrix{Float64}(undef, T_valid, size(states, 1))
```
MartinuzziFrancesco: don't force Float64 if you can infer it from input/states

lib/ReservoirComputingBenchmarks/src/ipc.jl
```diff
@@ -0,0 +1,215 @@
...
+        cross_terms::Bool = true,
+        train_ratio::Real = 0.8,
+        reg::Real = 1.0,
+    )
+    T = length(input)
+    n_features = size(states, 1)
+    @assert size(states, 2) == T "states must have $T columns (time steps), got $(size(states, 2))"
+    @assert max_delay >= 1 "max_delay must be >= 1, got $max_delay"
+    @assert max_degree >= 1 "max_degree must be >= 1, got $max_degree"
+    @assert max_delay < T "max_delay ($max_delay) must be less than signal length ($T)"
```
MartinuzziFrancesco: instead of assertions I would make more use of throw(ArgumentError(...)) or DimensionMismatch(...). making clear in the docs what the dims should be and then throwing the errors should be a cleaner path. I know I'm using them a bunch in the main library, but that should change too

MartinuzziFrancesco:
I would say that the library is in a good stage and ready to be merged. I do have a couple of comments, which I submitted as a review but I see them as more general, since they appear in more functions:

 - I think having @assert s is not the way to go, and probably throwing specific errors (when possible) would be clearer for the user
 - try to avoid forcing Float64 s and try to infer the type from the input data / states
 - I'd leave the functions (like `metric`) untyped

Saswatsusmoy:
Hi @MartinuzziFrancesco — pushed three commits (`09985d2`, `2def5e2`, `4b87049`) addressing the three themes from your review:

1. `refactor(benchmarks): replace @assert with ArgumentError/DimensionMismatch` — argument-value violations now throw `ArgumentError`, shape/length mismatches throw `DimensionMismatch`. Tests updated accordingly.
2. `refactor(benchmarks): infer element type from inputs instead of forcing Float64` — each benchmark derives a single `Telt = promote_type(eltype(input), eltype(states), typeof(reg))` and uses it for every internal buffer; `generate_narma` promotes across the input + four coefficients. Metric and `_normalize` helpers made type-preserving so storage no longer silently downcasts. Docstrings updated.
3. `refactor(benchmarks): drop MetricFunction alias; accept any callable metric` — `::MetricFunction` annotation removed from `narma`, unused `const MetricFunction` deleted from the module.

All 180 tests still pass, Runic format is clean.

### Disclosure
Full transparency: these three commits were drafted with an AI agent harness (Claude Code) rather than written by hand. I scoped the work this way because the changes are mechanical and low-risk — pure refactors that don't alter benchmark semantics, fully covered by the existing test suite, and easy to verify by reading the diff. I reviewed each commit before pushing, but flagging it explicitly so you can apply whatever extra scrutiny you'd like. Happy to redo any piece manually if you'd prefer.

MartinuzziFrancesco:
looks good to me, I think we may need @ChrisRackauckas s input on how to structure the docs for v 0.1, since this is a multi library setup. It can just be api documentation for this version, but I'd prefer that we take care of it in this pr

PR #421, opened 2026-04-15
Add MemoryResESNCell/RMNResESN residual reservoir memory network (#418)

### LINKED ISSUE #418: Residual Reservoir Memory Networks
## Paper

https://doi.org/10.1109/IJCNN64981.2025.11227859

https://arxiv.org/pdf/2508.09925

## Equations

Identical to RMN (#339) but with a ResESN (#417 ) for the nonlinear reservoir. The cells add a memory component.

## Official Implementation

https://github.com/NennoMP/residualrmn

Saswatsusmoy:
Closes #418.

Implements Ceni, Pintore & Gallicchio, *Residual Reservoir Memory Networks*, IJCNN 2025 (doi:10.1109/IJCNN64981.2025.11227859), built on top of the RMN family introduced in #354. Following the family-naming convention discussed in the original review thread, the cell is `MemoryResESNCell` and the model `RMNResESN`; the paper-side name *ResRMN* lives in the docstring `# Reference` line and bib entry.

## Equations

```math
\begin{aligned}
    \mathbf{m}(t) &= \mathbf{W}_{\text{in}}^{m}\, \mathbf{u}(t) + \mathbf{C}\, \mathbf{m}(t-1) + \mathbf{b}^{m} \\
    \mathbf{h}(t) &= \alpha\, \mathbf{O}\, \mathbf{h}(t-1) + \beta\, \phi\!\left(\mathbf{W}_{\text{in}}\, \mathbf{u}(t) + \mathbf{W}_r\, \mathbf{h}(t-1) + \mathbf{W}_m\, \mathbf{m}(t-1) + \mathbf{b}\right)
\end{aligned}
```

`C` defaults to a cyclic (shift) orthogonal matrix via `simple_cycle(; cycle_weight=1)`, matching the official PyTorch reference implementation (https://github.com/NennoMP/residualrmn).

## Added

- `MemoryResESNCell` (`src/layers/memoryresesn_cell.jl`): ResESN-style residual cell augmented with a memory-to-reservoir kernel `W_m`. Reduces exactly to `ResESNCell` when `W_m = 0` (verified in tests). Forward signature `(inp, (h, m))` matches `MemoryESNCell` so it drops into `RMNCell` as the nonlinear reservoir.
- `RMNResESN` (`src/models/rmnresesn.jl`): thin constructor mirroring `RMNESN`. Wires `ESNCell(in_dims => mem_dims, identity, init_reservoir = simple_cycle(; cycle_weight=1))` as the linear memory reservoir and `MemoryResESNCell` as the nonlinear reservoir into an `RMNCell` → `StatefulLayer` → modifiers → `LinearReadout` pipeline. No custom `_partial_apply`/`collectstates`/`resetcarry!` overrides — `RMNCell` handles dual-state plumbing.
- Tests in `test/layers/test_memoryresesncell.jl` (cell semantics, batch support, `W_m = 0` ≡ `ResESNCell` equivalence) and `RMNResESN` added to the `ESN-family model contract` loop in `test/models/test_esn.jl` mirroring the `RMNESN` dispatch.
- Bibliography entry `Ceni2025b`; docs entries in `docs/src/api/layers.md` and `docs/src/api/models.md`.

## Notes vs. previous revision

- Earlier revisions of this PR introduced a separate `MemoryCell` layer and a wrapper `ResRMN` model. Both are gone: the linear memory reservoir is now `ESNCell(identity)` (per @MartinuzziFrancesco's review), and the composite model dispatches through the standard `RMNCell` plumbing landed in #354.
- Resolves the Copilot docstring comments about `_default_memory_reservoir` (the helper no longer exists; the default is `simple_cycle(; cycle_weight=1)` as documented in the signature).

## Checklist


src/layers/memory_cell.jl
```diff
@@ -0,0 +1,109 @@
...
+
+  - `input_matrix :: (out_dims × in_dims)` — `W_in`
+  - `reservoir_matrix :: (out_dims × out_dims)` — `C`
+  - `bias :: (out_dims,)` — present only if `use_bias=true`
+
+## States
+
+  - `rng`: a replicated RNG used to sample initial hidden states when needed.
+"""
+@concrete struct MemoryCell <: AbstractEchoStateNetworkCell
```
MartinuzziFrancesco: The whole `MemoryCell` is just ESNCell with the identity function as activation right? I think you can remove this code and let the user plug `ESNCell(in_dims => out_dims, identity)` to the `RMN` of #354, just put it in the `ResRMN` call instead of the memory cell

src/layers/memory_cell.jl
```diff
@@ -0,0 +1,109 @@
...
+    out_dims <: IntegerType
+    init_bias
+    init_reservoir
+    init_input
+    init_state
+    use_bias <: StaticBool
+end
+
+_default_memory_reservoir(rng::AbstractRNG, dims::Integer...) =
+    simple_cycle(rng, dims...; cycle_weight = 1.0f0)
```
MartinuzziFrancesco: you can just call `init_reservoir = simple_cycle(; cycle_weight)`, no need to do this.

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
I like the design of the `ResRMN`, but I think that there is a bit of overengineering for a couple of things. I will test it a bit and report a couple of other comments

MartinuzziFrancesco:
I think I will need to finish up #354 before this one can be finalized, because the structure that I started to set up in that PR would remove a bunch of code here and make the implementation smoother. But in order to finalize that PR, and this one, I want to fix nomenclature for this family of models. This should help me finish the original PR, since I got stuck in this:

 - The reservoir memory network is a general concept: using a linear and nonlinear reservoirs to keep memory of the input. So the `RMNCell`, which takes in a `nonlinear_reservoir` and a `linear_reservoir` as defined in #354 is correctly named and currently does the correct thing. I think I will need to check if the size of the memory can be different from the size of the reservoir in the current implementation, but that should be the case.
 - From that it follows that the `RMN` model should not really exist: once the user defines the `RMNCell` they can just plug it into a `StatefulLayer` and then into a `ReservoirComputer`. The pieces that we need to provide to close this PR and #354 are the memory enhanced `MemoryESNCell` and `MemoryResESNCell`, which are both correctly implemented. In this point I would suggest that you rename `ResRMNCell` to be `MemoryResESNCell` so we have a standard for naming.
 - The final model would then borrow the naming from the nonlinear reservoir driver. So in #354 the name is wrong, and it should be something like either `RMNESN` or `ESNRMN`. In your case `ResRMN` could be a fine naming, since it's the one they use in the paper, but I would suggest the more specific `ResESNRMN` or `RMNResESN`, depending on which standard we decide. Let me know which one sounds better to you. I am partial for `RMN<cell_name>` since it makes them easier to identify.

This should be all, let me know your thoughts on this! After that I will finalize my PR and from it this one can follow as well

Saswatsusmoy:
Yeah this all makes sense, and I think you're right that the wrapper model is doing too much. Looking at `src/models/resrmn.jl` again, most of the 270 LOC is the `_partial_apply`/`collectstates`/`resetcarry!` overrides to keep the two carries in sync if `RMNCell` from #354 owns both reservoirs internally, all of that just collapses into the standard `StatefulLayer` path. So +1 to dropping `ResRMN` as a distinct model.                                                                                                              
   
On the naming, I'd go with **`RMN<cell>`** so `MemoryResESNCell` for the cell and `RMNResESN` if we keep a convenience constructor. Honestly the deciding factor for me is that it lines up with `DeepESN` / `HybridESN` already in the package modifier prefix, base reservoir on the right. The cell side you proposed (`Memory<base>`) reads the same way, so the whole family stays internally consistent. `ResESNRMN` would work but it sticks out next to the existing names.                                                                 
                                                                      
Paper calls it "ResRMN" but I don't think that's worth fighting for in the type name bib entry and a `# Reference` line in the docstring keep the paper link discoverable, and anyone grepping for "ResRMN" will hit the docs.
                                                                                                                                                                                     
One thing worth confirming on your side while you're in #354: the current `ResRMNCell` here already keeps `mem_dims` separate from `out_dims` (memory reservoir is its own size), so as long as `RMNCell` exposes that the abstraction carries over cleanly. I think you suspected this was already the case just flagging that it is, on this side at least.
                                                                                                                                                                                     
I'll sit on this branch until #354 lands. Once it's in I'll rebase, drop the wrapper, rename the cell to`MemoryResESNCell`, and either add `RMNResESN` as a thin constructor or skip it depending on what shape #354 settles on.

MartinuzziFrancesco:
Good point on the naming! I have somewhat of a working version of #354 where I still have to add docs and tests which should be in line with what we discussed. I'll try to finalize it and merge it in the coming days, unless I find any bug with the implementation

MartinuzziFrancesco:
all right, #354 is closed, so if you want this one can be finalized as well now

Saswatsusmoy:
Branch rebuilt on top of master now that #354 has landed. Summary of what changed vs. the previous revision:

- Dropped `MemoryCell` entirely — the linear memory reservoir is now `ESNCell(in_dims => mem_dims, identity, init_reservoir = simple_cycle(; cycle_weight = 1))`, plugged into `RMNCell` exactly the way `RMNESN` does it. Resolves [this thread](https://github.com/SciML/ReservoirComputing.jl/pull/421#discussion_r3092082638).
- Dropped the `ResRMN` wrapper model (270 LOC of `_partial_apply`/`collectstates`/`resetcarry!` overrides). `RMNCell` from #354 handles dual-state plumbing so the composite is now a thin constructor mirroring `RMNESN`.
- Renamed `ResRMNCell` → `MemoryResESNCell` and `ResRMN` → `RMNResESN` per the family-naming convention we agreed on. The paper-side name *ResRMN* lives in the docstring `# Reference` line and bib (`Ceni2025b`) for discoverability.
- `MemoryResESNCell` uses the partial-application initializer form `simple_cycle(; cycle_weight = 1)` for the memory recurrent kernel, matching the rest of the package. Resolves [this thread](https://github.com/SciML/ReservoirComputing.jl/pull/421#discussion_r3092118535) and the Copilot docstring comments (the private `_default_memory_reservoir` helper no longer exists).
- Tests: cell semantics in `test/layers/test_memoryresesncell.jl` (including the `W_m = 0 ⇒ ResESNCell` equivalence), `RMNResESN` added to the `ESN-family model contract` loop in `test/models/test_esn.jl` mirroring the `RMNESN` dispatch.

MartinuzziFrancesco:
this is looking clean to me, feel free to up the version to 0.12. 23 and we can merge it!

Issue #268, opened 2025-03-17
Sparse reservoir topologies


MartinuzziFrancesco:
## Description

Introduced in [Cossu2025](https://dl.acm.org/doi/10.1145/3703412.3703413), these topologies are built to increase performance while reducing complexity and randomness in reservoir initializations. They provide alternative constructions, some following the minimal complexity esn approach, others being created using a deterministic alternative to the usual `rand_sparse` approach. I believe all of these can implemented using the building blocks introduced in #267.

## Description

 -  **Lower triangular** (`lower_triangular()`): sampled uniformly in (-1, 1) and scaled. It provides weights on the diagonal and "lower" diagonals until a given sparsity is reached. This can be built stacking `delay_line!`s and a `scale_radius!` at the end.
 - **Band** (`band_init()`): sampled uniformly in (-1, 1) and scaled. Elements on the diagonal, and upper and lower diagonals until a given sparsity is reached. This is a `self_loop!` plus however many `delay_line!`s and `backward_connection!`s are needed. Closing with a `scale_radius!`
 - **Ring**: Essentially `simple_cycle`, rescaled to a chosen spectral radius. Stack `simple_cycle` and `scale_radius!`.
 - **Toepliz** (`toepliz_init`): essentially `band_init` but for each diagonal the elements are the same (think minimal complexity esn).
 - **Circulant** (`circulant_init`): Essentially `simple_cycle`, but with added diagonals below the top right element and the lower diagonal. This would be easy to obtain if we generalize the `simple_cycle!` function.

am-1118:
Hi @MartinuzziFrancesco, I'd love to take this up. I recently worked on the Wigner matrix initialisers, where I got a good idea of coding style and documenting. I'll read the paper and check the building blocks in #267. In case anyone else is working on this, please let me know.

PR #439, opened 2026-05-13
feat: implement band reservoir topology 

am-1118:
## Checklist

  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular  the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and [COLPRAC](https://github.com/SciML/COLPRAC).
  
## Additional context
This addition addresses Issue #268 by implementing the band topology introduced in [Cossu et al., 2025](https://dl.acm.org/doi/10.1145/3703412.3703413). 

This utilises the generic initialisation building blocks (`self_loop!`, `delay_line!`, etc.) recently merged in #267 to construct the matrices in-place.

### 🚧 Function Implemented
- [x] `band_init`

### 💡 Implementation Notes for `band_topology`
**Exact Sparsity:** To keep the matrix physically symmetrical, the function builds outward from the centre by adding matched pairs of upper and lower diagonals. To hit the exact user-requested `sparsity` target, the final outermost diagonals may not need to be full. Instead of filling these diagonals sequentially (which breaks the symmetry), the function calculates exactly how many weights are left and randomly allots them across those outer diagonals using `randperm`.

**Performance:** Implemented Compile-Time Value Dispatch (`Val(return_sparse)`) to ensure type stability when returning `SparseMatrixCSC` vs `Matrix`.

### DIFF
--- a/src/inits/inits_reservoir.jl
+++ b/src/inits/inits_reservoir.jl
@@ -2843,7 +2843,7 @@ function lower_triangular(
 end
 
 @doc raw"""
-    band_topology([rng], [T], dims...; radius=1.0, sparsity=0.9, return_sparse=false)
+    band_init([rng], [T], dims...; radius=1.0, sparsity=0.9, return_sparse=false)
 
 Create and return a sparse reservoir matrix with a band topology [Cossu2025](@cite).
 This function populates the main diagonal (self-loops) and iteratively expands outward to the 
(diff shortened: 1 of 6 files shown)


src/inits/inits_reservoir.jl
```diff
@@ -2842,6 +2842,125 @@ function lower_triangular(
...
+
+  - `rng`: Random number generator. Default is `Utils.default_rng()` from
+    [WeightInitializers](https://lux.csail.mit.edu/stable/api/Building_Blocks/WeightInitializers).
+  - `T`: Type of the elements in the reservoir matrix. Default is `Float32`.
+  - `dims`: Dimensions of the reservoir matrix. Must be square.
+
+## Keyword arguments
+
+  - `radius`: The desired spectral radius of the reservoir. Defaults to 1.0.
+  - `sparsity`: The exact target fraction of zero elements in the matrix.
```
MartinuzziFrancesco: I wouldn't say exactly, since there is a margin of approximation

src/inits/inits_reservoir.jl
```diff
@@ -2842,6 +2842,125 @@ function lower_triangular(
...
+  0.0      0.0      1.412   0.954
+```
+"""
+function band_topology(
+        rng::AbstractRNG, ::Type{T}, dims::Integer...;
+        radius = T(1.0), sparsity = 0.9, return_sparse = false
+    ) where {T <: Number}
+
+    throw_sparse_error(return_sparse)
+    check_res_size(dims...)
```
MartinuzziFrancesco, line 2917: also add a check for sparsity in [0, 1]

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
I would change the docs wording a bit, there is a bias for filling one diagonal over the other right?

src/inits/inits_reservoir.jl
am-1118: done, added an if-else block

src/inits/inits_reservoir.jl
am-1118: updated the docstring

PR #450, opened 2026-06-07
feat: continuous-time reservoirs via RCODEReservoirExt

Saswatsusmoy:
## Summary

Implements the second deliverable of the SciML 2026 Summer Fellowship (#397) — the `RCODEReservoirExt` package extension that makes `SciMLProblemReservoir` actually run. Builds on the core types from #446.

- Continuous `_collectstates` pipeline: interpolated input signal → `remake` → `solve(saveat = sample_grid)` → sampler.
- Teacher-forced and autoregressive `predict` overrides for `AbstractSciMLProblemReservoir`.
- Hand-rolled `LinearInputInterp` sidesteps a DataInterpolations matrix-`u` / SciMLBase dual-eltype probing interaction that crashes the bare `LinearInterpolation` path inside an ODE parameter pack.
- Construction-time rejection of the three protected `solve` kwargs (`saveat`, `save_everystep`, `dense`).
- Loud errors on the reserved `:input` parameter-name collision, on degenerate `tspan`, and on non-NamedTuple `prob.p`.

`states[:, k]` is the reservoir state after processing input column `k` — inputs are placed at window starts, samples are taken at window ends. With that alignment the Euler-equivalence check holds without an off-by-one shift.

## Design references

- Locked design from #397 (issuecomment-4595213663 and -4604259768).
- Follows the DiffEqFlux NeuralDE pattern for `args`/`kwargs`/`tspan` capture (per @MartinuzziFrancesco's sketch).
- Two-level `predict` -> `_predict` dispatch mirrors the `collectstates` pattern landed in #446.

## Test plan

Local Core suite (Julia 1.12.5): **1413/1413 pass**, including 34 new extension tests covering

- [x] Linear ODE analytic match (`dx/dt = -x + u, x(0) = 0`).
- [x] Euler equivalence with the discrete reservoir update (no off-by-one).
- [x] Sampler shape contract.
- [x] Teacher-forced and autoregressive `predict` (shape + determinism).
- [x] State modifiers compose with the continuous path.
- [x] `T_steps` and `steps` boundary errors.
- [x] Protected solve kwargs rejected at construction.
- [x] `tspan[2] > tspan[1]` enforced.
- [x] `:input` parameter-name collision errors.
- [x] `prob.p` accepts NamedTuple / `nothing` / `NullParameters`; rejects other types.
- [x] Aqua / ExplicitImports / JET / Runic all clean locally.

Roadmap for PRs 3–6 unchanged: CTESN model + Mackey-Glass/Lorenz validation (PR3), perf (PR4), LSM (PR5), DDE stretch (PR6).


Saswatsusmoy:
Pushed `9ef60c4a` addressing review feedback. Triage:

**Fixed**
- Switched the input signal to zero-order hold (`ZeroOrderHoldInterp`). Linear interpolation made `states[:, k]` depend on `data[:, k+1]` for any non-Euler solver, which contradicted the documented "state after processing input `k`" semantics. ZOH makes only `data[:, k]` influence `states[:, k]` regardless of solver, and aligns with the autoregressive `predict` path which already used ZOH. The call returns a `view`, so the input function is allocation-free in the ODE hot path.
- Dropped `collect(res.prob.u0)` and `collect(initialdata)` in autoregressive `predict` — these destroyed `SVector` / `ComponentArray` state types and errored on scalar `u0`. The loop only reads, never mutates in place.
- Autoregressive `predict` now allocates `output` from the first readout result (`similar(y, length(y), steps)`) so a readout returning a different eltype doesn't silently convert.
- Replaced `@assert` with `ArgumentError` in the new `_predict` fallback for consistency with the rest of the continuous-path API.
- Tests: dropped the `global Win_noparams` in favour of a `let`-block closure.

**Deferred (with reasoning)**
- The remaining allocation in `ZeroOrderHoldInterp`'s interior branch — sidestepped by the ZOH switch since the new call returns a view rather than a fresh vector; nothing further to do here.
- Discrete `predict(::AbstractLuxLayer, steps::Integer, …)` allocating `output` from `eltype(initialdata)` — pre-existing behaviour in `src/predict.jl`. Out of scope for this PR; flagging it for a follow-up so the discrete path can mirror the eltype-from-readout pattern uniformly.

Local suite still green: 34/34 extension, 1413/1413 Core, Aqua 10/10, ExplicitImports 2/2, JET clean, Runic clean.

ext/RCODEReservoirExt.jl
```diff
@@ -0,0 +1,387 @@
...
+    data::D
+    ts::T
+end
+
+function (interp::ZeroOrderHoldInterp)(t)
+    ts = interp.ts
+    n = length(ts)
+    t < ts[1] && return view(interp.data, :, 1)
+    t ≥ ts[end] && return view(interp.data, :, n)
+    k = searchsortedlast(ts, t)
```
MartinuzziFrancesco: where possible avoid single letter variables

MartinuzziFrancesco: this is valid for the whole pr

ext/RCODEReservoirExt.jl
```diff
@@ -0,0 +1,387 @@
...
+    col1, new_st = ReservoirComputing._apply_seq(
+        modifiers, view(states_matrix, :, 1), ps_mods, st_mods
+    )
+    # `similar(col1, ...)` — not `similar(states_matrix, ...)` — so the
+    # output matrix takes the modifier output's eltype. If a modifier
+    # promotes/demotes (e.g. Float32 → Float64), we want that to surface,
+    # not be silently truncated back to the reservoir state's eltype.
+    out = similar(col1, length(col1), T)
+    out[:, 1] .= col1
+    for t in 2:T
```
MartinuzziFrancesco: is there a way to use built-in functions like `eachcol`? if so that would be preferred here

MartinuzziFrancesco: this is valid for most `for` loops

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
beside a couple of comments this looks good to me. I think that having a docs page would also help the user, I forgot if this is a later deliverable or if it should go in here

MartinuzziFrancesco:
everything looks good, after we fix the conflicts I think this is good to merge. Could you provide a quick eye test and reproduce the readme example with a continuous esn? just feeding the raw equations into the `SciMLProblemReservoir`, no need to do a proper `CTESN` struct here

MartinuzziFrancesco:
Sorry for my delay on this, everything looks good! up the version and we can merge this

Saswatsusmoy:
The GPU doc CI is failing as docs/Project.toml pins ReservoirComputing = "0.12.0" but main bumped to 0.13.0. what should I do ? widen docs compat to "0.12, 0.13" (or just "0.13"). Also I bumped up to 0.13.0 due to major changes can bump it from 0.12.23 -> 0.12.24. Which one seems apt here?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ReservoirComputing"
 uuid = "7c2d2b1e-3dd4-11ea-355a-8f6a8116e294"
-version = "0.12.23"
+version = "0.13.0"
```
MartinuzziFrancesco: sorry I should've specified, it should be 0.12.24 as I'm keeping 0.x for breaking changes until v1.0

MartinuzziFrancesco (changes requested):

PR #459, opened 2026-06-22
fix: removing delay and nla bugs, small all around fixes

MartinuzziFrancesco:
## Checklist


src/models/esn_delay.jl
```diff
@@ -517,3 +517,42 @@ function Base.show(io::IO, esn::DelayESN)
...
     return
 end
+
+# `InputDelayESN`/`DelayESN` carry an extra `input_delay` field that must run
+# *before* the reservoir. The generic `AbstractReservoirComputer` machinery in
+# `reservoircomputer.jl` only knows about `(:reservoir, :states_modifiers,
+# :readout)`, so these models need their own parameter/state setup and their
+# own `_partial_apply`. The generic forward call and `collectstates` both route
+# through `_partial_apply`, so overriding it here is enough to make them work.
+const _InputDelayedESN = Union{InputDelayESN, DelayESN}
```
Saswatsusmoy, line 527: Curious about one thing: since `AbstractReservoirComputer{Fields}` already encodes the field-name tuple, would it make sense down the line to have the generic `initialparameters` / `_partial_apply` introspect `Fields` directly, so any future composite model with extra pre- or post-fields just works? Could be a follow-up — wondering whether you'd considered that direction, or whether there's a reason a per-model override is preferable here.

src/states.jl
```diff
@@ -375,21 +375,21 @@ julia> mat_old = [1  2  3;
...
+ 208  238  270
 
 ```
 """
 function NLAT2(x_old::AbstractVector)
     x_new = copy(x_old)
     for idx in eachindex(x_old)
-        if firstindex(x_old) < idx < lastindex(x_old) && isodd(idx)
-            x_new[idx, :] .= x_old[idx - 1, :] .* x_old[idx - 2, :]
+        if firstindex(x_old) < idx && isodd(idx)
```
Saswatsusmoy, line 391: When `firstindex(x_old) > 1` (e.g. an `OffsetArray`), this would start applying at `idx = firstindex + 1` if odd. Is that the intended semantics for offset inputs, or is NLAT2 effectively assumed to receive standard 1-based vectors? Probably never comes up in practice but I noticed it.

test/test_esn_delay.jl
```diff
@@ -188,3 +188,119 @@ end
...
+# guard against the `input_delay` field being silently dropped from params/state
+# or skipped during application (which used to crash with a `DimensionMismatch`).
+@testset "InputDelayESN forward pass & training" begin
+    rng = MersenneTwister(123)
+    in_dims, res_dims, out_dims, num_delays = 3, 40, 3, 2
+    n_steps = 60
+    data = rand(rng, Float32, in_dims, n_steps)
+    target = rand(rng, Float32, out_dims, n_steps)
+
+    idesn = InputDelayESN(in_dims, res_dims, out_dims; num_delays = num_delays)
```
Saswatsusmoy, line 199: None of the new round-trip tests touch `stride > 1`. The fix itself isn't stride-sensitive (stride only changes what the buffer keeps), but I wondered if a quick `stride=2` smoke test might be worth adding here just to lock the end-to-end behavior down. Or do you think the current coverage is enough?

Saswatsusmoy (commented):
Read through this carefully. The `InputDelayESN`/`DelayESN` one looks like it had been silently broken since the input-delay split, since the constructor tests never ran a forward pass — the `newst.input_delay.clock == n_steps` assertion is a useful regression guard. Left a few questions inline, all curiosity, nothing blocking.

src/models/esn_delay.jl
MartinuzziFrancesco: that could be an elegant solution yeha, if  there's a way to make it generic enough it should solve similar gotchas to this one

src/states.jl
MartinuzziFrancesco: This isn't the only algo that would have issues with nonstandard arrays. It's in the plans to make a larger check for this kind of issues

test/test_esn_delay.jl
MartinuzziFrancesco: yeha that would be a good idea, I'll add it

PR #470, opened 2026-07-13
Refactor DeepESN to accept arbitrary Lux layers (#375)

### LINKED ISSUE #375: Extend deep models to take any cell/layer with `DeepReservoir`
Currently the library supports the fully built model `DeepESN`. However, as the name suggests, the model has the `ESNCell` hardcoded as reservoirs. Providing a `DeepReservoir` model with the same general behaviour would allow to explore other architectures with deep connetions

am-1118:
## Checklist

- [ ] Appropriate tests were added

## Additional context

**Resolves #375**

This PR refactors `DeepESN` to act as a generic wrapper that accepts a `Tuple` of arbitrary layers (e.g., standard `ESNCell`s, custom dynamical cells, or standard Lux Dense layers), rather than hardcoding `ESNCell` instantiation via a `res_dims` integer array.

**Key Architectural Changes:**
* **Universal Wrapper Pattern:** Replaced the `res_dims` array requirement with a generic `Tuple` of cells. Arbitrary cells are now passed directly into the constructor and wrapped in `StatefulLayer` automatically if needed (mirroring the `LocalInformationFlow` design).
* **Type Stability & Performance:** Rewrote the forward pass (`_partial_apply`) and `collectstates`. Completely removed the type-unstable `Vector{Any}` loops, replacing them with strict, type-stable `ntuple` operations. This ensures zero-overhead performance inference even when users stack completely different types of custom layers.
* **Lux Integration:** Leans fully into Lux's explicit state paradigm, natively routing parameters (`ps`) and continuous reservoir memory (`st`) down the deep chain regardless of the underlying cell type.

### DIFF
--- a/src/models/esn_deep.jl
+++ b/src/models/esn_deep.jl
@@ -96,57 +96,20 @@ Composition:
 end
 
 function DeepESN(
-        in_dims::IntegerType,
-        res_dims::AbstractVector{<:IntegerType},
-        out_dims::IntegerType,
-        activation = tanh;
-        leak_coefficient = 1.0,
-        init_reservoir = rand_sparse,
-        init_input = scaled_rand,
-        init_bias = zeros32,
-        init_state = randn32,
-        use_bias = false,
-        state_modifiers = (),
-        readout_activation = identity
+        cells::Tuple,
+        readout;
+        states_modifiers = nothing
     )
-    n_layers = length(res_dims)
-    acts = _asvec(activation, n_layers)
-    leaks = _asvec(leak_coefficient, n_layers)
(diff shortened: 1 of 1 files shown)


MartinuzziFrancesco:
The idea behind #375 is to have an additional layer that generalizes. The example I mentioned is the LIF model, whicvh has both the wrapper and the original LIFESN model. So the goal should not be a refactor of DeepESN here, but the addition of a DeepReservoir wrapper

am-1118:
I will close this refactoring PR, revert the changes to DeepESN, and open a fresh PR that adds the new DeepReservoir wrapper alongside it as an additional feature. Thanks for the clarification!

PR #471, opened 2026-07-13
Add DeepReservoir universal wrapper for arbitrary Lux layers (#375)

### LINKED ISSUE #375: Extend deep models to take any cell/layer with `DeepReservoir`

am-1118:
## Checklist

## Additional context

**Resolves #375**

This PR introduces a new `DeepReservoir` architecture to serve as a universal layer wrapper.

**Key Additions:**
* **`DeepReservoir` Wrapper:** Added a new struct and constructor that accepts a `Tuple` of arbitrary `AbstractExplicitLayer`s (e.g., standard `ESNCell`s, custom dynamical cells, or standard Lux Dense layers). 
* **Backward Compatibility:** The original `DeepESN` struct and its array-based integer constructors remain completely untouched to preserve the existing API for legacy users.
* **Type-Stable Forward Pass:** Implemented `_partial_apply` and `collectstates` for the new wrapper using strict, type-stable `ntuple` operations to ensure zero-overhead performance inference across hybrid layer stacks.
* **Lux Integration:** `DeepReservoir` leverages Lux's explicit state paradigm, natively routing parameters (`ps`) and continuous reservoir memory (`st`) down the deep chain regardless of the underlying cell type.


src/models/esn_deep.jl
```diff
@@ -293,3 +289,240 @@ end
...
+
+function DeepReservoir(
+        cells::Tuple,
+        readout;
+        states_modifiers = nothing
+    )
+    n_layers = length(cells)
+
+    # Ensure cells are wrapped in StatefulLayer to maintain continuous memory
+    stateful_cells = map(c -> c isa StatefulLayer ? c : StatefulLayer(c), cells)
```
MartinuzziFrancesco: I'm  not too sure about hard coding this, there might be layers that are not necessarily deep. Maybe you can add a boolean toggle for this to make life easier for the users

src/models/esn_deep.jl
```diff
@@ -293,3 +289,240 @@ end
...
+            cell_out, st_cell_i = apply(desn.cells[idx], current_inp, ps.cells[idx], newst.cells[idx])
+
+            mod_out, st_mods_i = _apply_seq(
+                desn.states_modifiers[idx], cell_out,
+                ps.states_modifiers[idx], newst.states_modifiers[idx]
+            )
+
+            current_inp = mod_out
+            return (cell = st_cell_i, mod = st_mods_i)
+        end
```
MartinuzziFrancesco: you have written _partial_apply for this. avoid code reuse

src/models/esn_deep.jl
```diff
@@ -293,3 +289,240 @@ end
...
+
+        push!(collected, copy(current_inp))
```
MartinuzziFrancesco: any specific reason for pushing a copy here?

src/models/esn_deep.jl
```diff
@@ -293,3 +289,240 @@ end
...
+    end
+
+    @assert !isempty(collected)
+    states = eltype(data).(reduce(hcat, collected))
+
+    return states, newst
+end
+
+function initialparameters(rng::AbstractRNG, dres::DeepReservoir)
+    ps_cells = map(l -> initialparameters(rng, l), dres.cells) |> Tuple
```
MartinuzziFrancesco: avoid single letter variables

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
Looking good so far, I have left a couple of comments. Also, put the implementation in a new file, deep_reservoir.jl in the same folder models

src/models/esn_deep.jl
am-1118: I've removed the duplicate ntuple loop block in collectstates and replaced it with a direct call to _partial_apply

src/models/esn_deep.jl
am-1118: It is redundant, removed it

src/models/esn_deep.jl
am-1118: Changed l -> layer

src/models/esn_deep.jl
am-1118: Added a make_stateful kwarg. It defaults to true (acting like the old behavior), but accepts either a single boolean or a Tuple of booleans so users have layer by layer control.

am-1118:
Do I need to add a separate test file to the test/models?

MartinuzziFrancesco:
yeha a separate file would be the way to go. I would try to have it loop on all cells as well, so we know that it is composable

src/models/deep_reservoir.jl
```diff
@@ -0,0 +1,223 @@
+@doc raw"""
+    DeepReservoir(cells::Tuple, readout; states_modifiers=nothing)
```
MartinuzziFrancesco: avoid types in the function signature, specify them in the args or kwargs docs

src/models/deep_reservoir.jl
```diff
@@ -0,0 +1,223 @@
...
+    \mathbf{x}^{(\ell)}(t) &= (1-\alpha_\ell)\, \mathbf{x}^{(\ell)}(t-1)
+        + \alpha_\ell\, \phi_\ell\!\left(\mathbf{W}^{(\ell)}_{\text{in}}\,
+        \mathbf{u}^{(\ell-1)}(t) + \mathbf{W}^{(\ell)}_r\, \mathbf{x}^{(\ell)}(t-1)
+        + \mathbf{b}^{(\ell)} \right), \quad \ell = 2,\dots,L, \\
+    \mathbf{u}^{(\ell)}(t) &= \mathrm{Mods}_\ell\!\left(\mathbf{x}^{(\ell)}(t)\right),
+        \quad \ell = 2,\dots,L, \\
+    \mathbf{y}(t) &= \rho\!\left(\mathbf{W}_{\text{out}}\, \mathbf{u}^{(L)}(t)
+        + \mathbf{b}_{\text{out}} \right).
+\end{aligned}
+```
```
MartinuzziFrancesco: I think we can remove this

src/models/deep_reservoir.jl
```diff
@@ -0,0 +1,223 @@
...
+  > Exact field names for modifiers/readout follow their respective layer definitions.
+
+## States
+
+  - `cells :: NTuple{L,NamedTuple}` — states for each cell in the sequence.
+  - `states_modifiers :: NTuple{L,Tuple}` — per-layer tuples of modifier states.
+  - `readout` — states for the readout layer.
+
+"""
+@concrete struct DeepReservoir <: AbstractEchoStateNetwork{(:cells, :states_modifiers, :readout)}
```
MartinuzziFrancesco: subtype as <: AbstractReservoirComputer

src/models/deep_reservoir.jl
```diff
@@ -0,0 +1,223 @@
...
+        out, partial_st = _partial_apply(desn, inp, ps, newst)
+        push!(collected, out)
+        newst = merge(partial_st, (readout = newst.readout,))
+    end
+
+    @assert !isempty(collected)
+    states = eltype(data).(reduce(hcat, collected))
+
+    return states, newst
+end
```
MartinuzziFrancesco: we shouldn't need this now, the defualt fallback will take care of the state collection

src/models/deep_reservoir.jl
```diff
@@ -0,0 +1,223 @@
...
+
+    st_ro = initialstates(rng, dres.readout)
+    return (cells = st_cells, states_modifiers = st_mods, readout = st_ro)
+end
+
+function (dres::DeepReservoir)(inp, ps, st)
+    out, new_st = _partial_apply(dres, inp, ps, st)
+    inp_t, st_ro = apply(dres.readout, out, ps.readout, st.readout)
+    return inp_t, merge(new_st, (readout = st_ro,))
+end
```
MartinuzziFrancesco: we don't need this, the [default fallback](https://github.com/SciML/ReservoirComputing.jl/blob/master/src/reservoircomputer.jl#L83-L87) should take care of the forward pass

MartinuzziFrancesco (changes requested):

src/models/deep_reservoir.jl
am-1118: Removed

src/models/deep_reservoir.jl
am-1118: Removed

src/models/deep_reservoir.jl
am-1118: Done

src/models/deep_reservoir.jl
am-1118: Removed

src/models/deep_reservoir.jl
am-1118: Removed "::Tuple"

MartinuzziFrancesco:
I think the main idea is good now, you should only fix the errors and we can merge this

PR #474, opened 2026-07-14
Make LinearSolve the default ridge solver and prefer train over train!

### LINKED ISSUE #367: `train!` should not have a bang
It doesn't modify anything in place

### LINKED ISSUE #473: Make LinearSolve.jl the default solver for the linear readout
At the moment `LinearRedout` uses a in-house ridge regression implementation, which is working fine. However, I think that having LinearSolve.jl handle the, well, linear solve of the ridge problem would be more robust.

This issue is linked to [this comment](https://github.com/SciML/ReservoirComputing.jl/issues/367#issuecomment-3778714603), so the implementation would be breaking.

I also do not think that removing LinearSolve from ext and having it by default would add too much in precompilation time

Saswatsusmoy:
## Checklist

- [ ] Any code changes were done in a way that does not break public API

## Summary

Implements [#473](https://github.com/SciML/ReservoirComputing.jl/issues/473) and the soft landing for [#367](https://github.com/SciML/ReservoirComputing.jl/issues/367).

### Public API

- Primary training entry point is non-mutating `train(rc, train_data, target_data, ps, st; objective, solver, washout, return_states)`.
- Ridge training separates **objective** (`StandardRidge`) from **solver** (default `QRFactorization()`, optional legacy `QRSolver()`, or other LinearSolve algorithms).
- `train!` remains as a **deprecated** compatibility wrapper (`Base.depwarn`) that maps positional `train_method` to `objective`. Removal is left for a future breaking release.
- `QRFactorization` is re-exported so the default path works with only `using ReservoirComputing`.

### LinearSolve / ridge

- LinearSolve.jl is a **hard dependency** (package extension `RCLinearSolveExt` removed; logic in `src/train.jl`).
- Default ridge path uses the **Tikhonov-augmented** least-squares system (same construction as legacy `QRSolver`), solved with multi-column RHS via LinearSolve — not normal equations.
- Compat: `LinearSolve = "4.2"` (multi-RHS support).
- LIBSVM backends reject an explicit `solver` keyword with a clear `ArgumentError`.

### Docs and tests

- Tutorials, README, and examples prefer `train`.
- Tests: ridge characterization, dual API / deprecation, MLJ and LIBSVM model-level edge cases.

### Version

- `Project.toml` version: **0.12.28**.

## Breaking changes

1. **Default ridge solver** is LinearSolve `QRFactorization()` (was in-house `QRSolver()` when `solver` was omitted). Results may change for some problems even though both target ridge.
2. **LinearSolve.jl is required** (no longer optional via package extension).
3. **LinearSolve compat requires 4.2+**.
4. **`train!` is deprecated** in favor of `train` (still works with a deprecation warning).

## Migration

```julia
# Before
ps, st = train!(model, data, targets, ps, st, StandardRidge(1e-3))

# After
ps, st = train(model, data, targets, ps, st; objective = StandardRidge(1e-3))
# optional: solver = QRSolver()  # legacy escape hatch
# optional: solver = SVDFactorization()  # needs using LinearSolve
```

## Test plan

- [x] `test/test_train_ridge.jl`
- [x] `test/test_train_api.jl`
- [x] `test/test_train_ext_backends.jl` (MLJ + LIBSVM)
- [x] Related model tests updated to `train` (delay ESNs, NGRC, LIFESN, SVESM)
- [ ] Full Core + QA CI on this PR

Closes #473
Closes #367


MartinuzziFrancesco:
This look good and I think we can merge after I make up my mind on a couple of naming conventions. In the meantime one direction would be to approach deprecations in a more systematic way, following [these suggestions](https://invenia.github.io/blog/2022/06/17/deprecating-in-julia/), namely moving the deprecations to deprecated.jl and test/deprecated.jl. This way when we move on to v1.0 we delete those files and the cleanup is done

docs/src/api/train.md
```diff
@@ -1,13 +1,18 @@
...
-    train!
     train
+    train!
+    QRSolver
+    QRFactorization
 ```
 
+For ridge regression, `solver = nothing` selects LinearSolve's
+[`QRFactorization`](@ref). Other LinearSolve algorithms require
+`using LinearSolve`.
```
MartinuzziFrancesco: what if we just reexport LinearSolve so that all the methods are already available when `using ReservoirComputing`? pros/cons? 

Also pinging @ChrisRackauckas for on opinion on this direction

docs/src/tutorials/getting_started.md
```diff
@@ -180,8 +180,9 @@ be done by passing the `washout` keyword argument to `train`.
 #define training method
 training_method = StandardRidge(0.0)
 
-ps, st = train!(esn, input_data, target_data, ps, st, training_method;
-    washout = 0 # we use no washout
+ps, st = train(esn, input_data, target_data, ps, st;
+    objective = training_method,
```
MartinuzziFrancesco: now this should not be called training method as a variable, in the new setup it can be confusing

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
I think this could be a good time to address also the fact that `StandardRidge` doesn't sound too good in the new `train` refactor. `RidgeRegression` would be more appropriate. I know that `MLJLinearModels` uses the same name, but now that `LienarSolve` is the default ridge solver I think that people will prioritize the rc.jl native implementation

Issue #467, opened 2026-07-10
Performance: continuous-time reservoirs (PR4 of #397)


Saswatsusmoy:
## Summary

Tracking issue for **performance work on continuous-time reservoirs** (`ContinuousESN` / `SciMLProblemReservoir` / `RCODEReservoirExt`), corresponding to **PR4** of the SciML fellowship plan on #397.

PR1–PR3 landed the correct continuous core (#446, #450, #456). This issue tracks data-driven optimizations so each change can land as a small, reviewable PR with a before/after measurement.

Profiling-only harness (no package code changes): `benchmarks/continuous_profile/` (local for now; can be proposed separately if useful upstream).

## Motivation

On Apple M4 / Julia 1.12.5 / single-thread OpenBLAS, with `reltol=1e-6`, `abstol=1e-8`, Float64, `use_bias=true`, seed 17:

| Workload | Setup | Median |
|----------|--------|--------|
| **W3** | `ContinuousESN` `collectstates`, N=500, T=1000, **Tsit5** | **~13.5 s** |
| **W4** | same, **Euler** `dt=1` | **~42 ms** |
| **W5 / W5t** | Rodas5 vs Tsit5, N=100, T=200 | **~1.57 s / ~0.12 s** |
| **W7** | AR `predict`, N=500, 250 steps, Tsit5 | **~0.76 s** |
| **W10b** | discrete `ESN` `collectstates`, N=500, T=1000 | **~42 ms** |

Key ratios:

- Tsit5 / Euler (matched N=500, T=1000): **~320×**
- Continuous Tsit5 / discrete ESN (N=500): **~350×**
- Rodas5 / Tsit5 (matched N=100, T=200): **~13×**

Stage timers show **`solve` ≈ 100% of wall time** on every continuous path. Grid construction, ZOH, NamedTuple merges, and modifiers are negligible once warm.

### 1. Adaptive stepping dominates, not the RHS wrapper

- W3 (Tsit5, N=500, T=1000): **330 537** RHS evaluations (`nf`), **21 262** accepted steps, **33 827** rejected.
- W4 (Euler, same problem): **1 001** RHS evaluations.
- Default RHS `_continuous_esn_rhs!` is already **0 allocations / call** (`mul!` + in-place broadcast). Re-optimizing matvec alone will not hit a 3× target if NFE counts stay the same.

### 2. Teacher-forced path is one long `solve`

[...]

PR #479, opened 2026-07-27
perf(ext): reuse ODE integrator across AR predict windows

Saswatsusmoy:
## Summary

First data-driven perf change on the continuous-time AR `predict` path, addressing part of #467 (PR4 of the #397 track).

Refactors `_predict` on both `ContinuousESNCell` and `AbstractSciMLProblemReservoir` so the `ODEIntegrator` is built once and reused across every autoregressive window via `reinit!`, instead of allocating a fresh `ODEProblem` / `ODEIntegrator` / `ConstantInterpolation` per step. The per-window input signal moves into a small mutable `ConstantInputWindow` whose single field the AR loop rebinds in place.

Windows remain independent solves (`erase_sol = true`, `reset_dt = true`) — the AR math (Lukoševičius 2012 §3.2.6 eq. (5) evaluated over `[t_lo, t_hi]` with `x(t_lo) = current_state` and constant `u(t) = current_input`) is preserved.

## What feature report was addressed?

Part of #467 (PR4 continuous-time perf work). Roadmap slot from #397.

## Data (this branch vs `master`, apples-to-apples)

15-sample BenchmarkTools microbench under the profiling env (SciMLBase v2.155, OrdinaryDiffEq v6.111, Julia 1.12.5, Apple M4, single-thread OpenBLAS). Workload: `N = 500`, 250 AR steps, `Tsit5(; reltol=1e-6, abstol=1e-8)`.

| Metric | `master` | this PR | Δ |
|---|---:|---:|---:|
| Wall median | 427.32 ms | 418.97 ms | -2.0% |
| Wall σ | 31.5 ms | 11.5 ms | **-63%** |
| Wall max | 505.59 ms | 449.95 ms | -11.0% |
| Memory | 10.68 MiB | 2.18 MiB | **-79.6%** |
| Allocs | 50,008 | 5,892 | **-88.2%** |
| GC (max) | 1.15% | 0.00% | eliminated |

Median wall is essentially flat (within noise). The concrete wins are **allocation** (-80% bytes, -88% alloc count), **variance** (σ cut by 63%, tail max cut by 11%), and **zero GC** in every sample. The wall speedup story stacks with #467's other levers (specialization, tolerance/solver defaults) — not carried by this PR alone.

Per-step arithmetic on the L1 numbers: `20.4 MB / 250 = 82 KB/step` → `4.4 MB / 250 = 18 KB/step` — matches the profiled ~76 KB/step `solve` + ~3 KB `ODEProblem` build + ~0.7 KB `ConstantInterpolation` per-step allocation eliminated.

## What changed

| Before (per AR step) | After |
|---|---|
| `ConstantInterpolation([u, u], [t_lo, t_hi]; cache_parameters=true)` | mutate `input_fn.u_vec = current_input` |
| `_build_solve_params(...)` NamedTuple | built once upfront |
| `remake(res.prob; tspan, p, u0)` → fresh `ODEProblem` | built once upfront |
| `solve(sub_prob, alg; saveat=[t_hi], ...)` → fresh integrator + sol | `reinit!(int, u; t0, tf, erase_sol=true, reset_dt=true)` + `solve!(int)` |
| `current_state = sol.u[end]` | `current_state = integrator.u` |

`ConstantInputWindow{T <: AbstractVector}` (mutable, single field `u_vec`) lives next to `_make_const_input_fn` in `ext/RCODEReservoirExt.jl`. Callable — `(f::ConstantInputWindow)(t) = f.u_vec` — so the RHS's `p.input(t)` contract is unchanged.

`_make_const_input_fn` / `ConstantInterpolation` are left in place; the extension no longer uses them internally but they are still reachable via the local profiling harness and removing them is a separate cleanup.

Also bundled: **`ODEProblem{true, FullSpecialize}` on the two cell-path constructions**. Sidesteps `AutoSpecialize`'s `FunctionWrappersWrapper` boxing of the RHS; on `master` these show up as 5 distinct alloc types in the W7 profile, with this PR they vanish. Only affects the cell path — `SciMLProblemReservoir`'s `prob` is user-supplied and keeps its own specialization. Standalone win vs `master` (measured on a sibling branch) was -38% allocs / -74% σ; the delta on top of the integrator reuse above is small (~1% allocs) because the wrappers were only ever created 250× on the pre-reuse path — but the change is directionally correct, keeps the two cell entry points consistent, and eliminates the `FunctionWrappersWrapper` types from the alloc profile.

## Design notes

- **Why `erase_sol=true, reset_dt=true`**: each AR window is an independent continuous-time solve with fresh `u0` and constant `u(t)` — no state must leak across windows, including the adaptive stepper's `dt` cache. Both flags reset the integrator to the same starting condition a fresh `solve()` would use.
- **Why a mutable input wrapper (not `remake` / new NamedTuple per step)**: `p::NamedTuple` field structure is immutable, but the object it holds isn't. Mutating `input_fn.u_vec` in place propagates through `integrator.p.input(t)` at zero per-step alloc without touching the integrator's type-specialized caches.
- **Non-mutating `u0` paths (`SVector`, `ComponentArray`)**: `current_state = integrator.u` aliases the integrator's internal `u` field, which for `iip = true` with `Vector{T}` is a single persistent buffer. Behaviour matches the previous `current_state = sol.u[end]` pattern for the common `Vector{Float64}` path; exotic `u0` types are covered by the same tests as before and still pass.

## Test plan

- [x] Full ext suite passes: **34 / 34 tests, 12 / 12 testsets**, 0 fails, 0 errors
  - `Autoregressive predict` (the direct target)
  - `Euler equivalence with discrete reservoir update` (guards eq. (5) math)
  - `Linear ODE analytic match` at `atol = 1e-6`
  - `Teacher-forced predict` (unchanged path, sanity check)
  - `State modifiers on continuous path`, `Boundary sizes`, `Protected solve kwargs rejected at construction`, `tspan strictly positive`, `Reserved :input key collision errors`, `prob.p accepts NamedTuple / nothing / NullParameters`, `Non-NamedTuple prob.p errors clearly`, `TerminalStateSampling output shape`
- [x] Runic clean
- [x] Local Aqua / ExplicitImports / JET clean
- [x] Micro-benchmark run 15 samples × 30 s budget under identical env (numbers table above)

## Checklist

- [x] Appropriate tests were added *(existing coverage exercises the refactored path end-to-end; no new tests were needed for a semantics-preserving change)*
- [x] All documentation related to code changes were updated *(no user-facing API change)*

## Files

| Path | Role |
| --- | --- |
| `ext/RCODEReservoirExt.jl` | `ConstantInputWindow`; `_predict` refactor (both methods) |
| `test/qa/Project.toml` | Declare `ExplicitImports` as a direct dep of the QA env (was reachable only as a transitive dep via `SciMLTesting`; a fresh `Pkg.instantiate()` + `using ExplicitImports` failed) |

- Remove `_make_const_input_fn` + the `ConstantInterpolation` / `DataInterpolations` weakdep once the local profiling harness is updated to construct its own input.
- Sparse `jac_prototype` when the reservoir matrix is sparse (implicit-solver path).

## Additional context

Follows the profiling in `benchmarks/continuous_profile/` (not part of this PR); the ranked backlog in that harness's `REPORT.md` names *"Reuse ODE integrator / remake across AR steps"* as the top predict-path lever. The `fix(qa)` commit is a one-line env fix that unblocks the local precommit hook — happy to split it into its own PR if preferred.

### DIFF
--- a/ext/RCODEReservoirExt.jl
+++ b/ext/RCODEReservoirExt.jl
@@ -4,7 +4,8 @@ using DataInterpolations: ConstantInterpolation
 using LinearAlgebra: mul!
 using LuxCore: apply
 using Random: AbstractRNG
-using SciMLBase: ODEProblem, init, reinit!, remake, solve, solve!, NullParameters
+using SciMLBase: FullSpecialize, ODEProblem, init, reinit!, remake, solve, solve!,
+    NullParameters
 
 using ReservoirComputing: ReservoirComputing,
     AbstractReservoirComputer,
@@ -398,7 +399,7 @@ function _collectstates(
     # state, the parameter pack, and the input signal share a numeric
     # type. The user controls eltype through the `init_*` initialisers.
     u0 = zeros(eltype(ps.reservoir.input_matrix), cell.out_dims)
-    prob = ODEProblem(cell.equations, u0, cell.tspan, solve_p)
+    prob = ODEProblem{true, FullSpecialize}(cell.equations, u0, cell.tspan, solve_p)
 
     sol = solve(
         prob, cell.args...;
@@ -450,7 +451,7 @@ function _predict(
 
     input_fn = ConstantInputWindow(current_input)
     solve_p = _build_solve_params(nothing, ps.reservoir, input_fn)
-    sub_prob = ODEProblem(
+    sub_prob = ODEProblem{true, FullSpecialize}(
         cell.equations, current_state, (window_starts[1], window_ends[1]), solve_p
     )
     integrator = init(
--- a/ext/RCODEReservoirExt.jl
+++ b/ext/RCODEReservoirExt.jl
@@ -278,7 +278,7 @@ function _predict(
     # conversion at the column assignment.
     local outputs
     for (step_idx, (t_lo, t_hi)) in enumerate(zip(window_starts, window_ends))
-        input_fn.u_vec = current_input
+        input_fn.u_vec = convert(typeof(input_fn.u_vec), current_input)
         reinit!(
             integrator, current_state;
             t0 = t_lo,
@@ -463,7 +463,7 @@ function _predict(
 
     local outputs
     for (step_idx, (t_lo, t_hi)) in enumerate(zip(window_starts, window_ends))
-        input_fn.u_vec = current_input
+        input_fn.u_vec = convert(typeof(input_fn.u_vec), current_input)
         reinit!(
             integrator, current_state;
             t0 = t_lo,
--- a/ext/RCODEReservoirExt.jl
+++ b/ext/RCODEReservoirExt.jl
@@ -4,7 +4,7 @@ using DataInterpolations: ConstantInterpolation
 using LinearAlgebra: mul!
 using LuxCore: apply
 using Random: AbstractRNG
-using SciMLBase: ODEProblem, remake, solve, NullParameters
+using SciMLBase: ODEProblem, init, reinit!, remake, solve, solve!, NullParameters
 
 using ReservoirComputing: ReservoirComputing,
     AbstractReservoirComputer,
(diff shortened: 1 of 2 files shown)


MartinuzziFrancesco:
this looks good, have you checked consistency of results before and after?

Saswatsusmoy:
Yes — ran an apples-to-apples parity check: same seed (17), same workload (`ContinuousESN` N=500, `Tsit5(; reltol=1e-6, abstol=1e-8)`, `collectstates` on a 3×500 input + 250-step `predict`). Master ext and this branch produce **bit-identical** output on both paths (`max abs diff = 0.0`, same `hash`). That's the strongest possible parity — no numerical drift, not even at the last-ULP level. `reinit!` with `erase_sol=true, reset_dt=true` genuinely reproduces the fresh-`solve` trajectory for these windows.

PR #480, opened 2026-07-28
perf(ext): sparse `jac_prototype` on the cell-path ODE

Saswatsusmoy:
## Summary

Second data-driven perf change on the continuous-time cell path (independent of #479), addressing part of #467 (PR4 of the #397 track).

Adds an opt-in `use_jac_prototype = false` kwarg on `ContinuousESNCell` and the `ContinuousESN` factory. When flipped `true`, the underlying `ODEFunction` is built with a sparse `jac_prototype` derived from the reservoir matrix — but only when a prototype is available for the current `(equations, W_r)` pair. Default (`false`) reproduces `master` bit-for-bit. Only affects the `ContinuousESNCell` path; `SciMLProblemReservoir` keeps its user-owned `prob` and specialization.

The prototype is `W_r + I` — exactly the sparsity of the analytic Jacobian

```math
\mathbf{J}(\mathbf{x}) \;=\; -\mathbf{I} \;+\; \mathrm{diag}\!\left(1 - \tanh^{2}(\mathbf{z})\right) \cdot \mathbf{W}_{r}
```

of the leaky-integrator ESN RHS (Lukoševičius 2012 §3.2.6 eq. (5)). Structure = `pattern(W_r) ∪ diag`; values in the prototype are ignored by the FD Jacobian, only the nonzero pattern matters.

## What feature report was addressed?

Part of #467 (PR4 continuous-time perf work). Roadmap slot from #397.

## API — user-facing toggle

```julia
model = ContinuousESN(
    in_dims, res_dims, out_dims, tspan, Rodas5();
    init_reservoir = (rng, d...) -> rand_sparse(rng, Float64, d...; return_sparse = true),
    use_jac_prototype = true,     # new — opt-in
    reltol = 1e-6, abstol = 1e-8,
)
```

Semantics matrix (all four cases verified locally):

| `use_jac_prototype` | `W_r` type | `equations` | resulting `jac_prototype` |
|---|---|---|---|
| `false` (default) | any | any | `nothing` (matches `master`) |
| `true` | `AbstractSparseMatrixCSC` | built-in `_continuous_esn_rhs!` | `W_r + I` (sparse) |
| `true` | dense | any | `nothing` (silent no-op) |
| `true` | sparse | user-supplied RHS | `nothing` (silent no-op — safe) |

The two silent no-op rows exist so the toggle is monotone: `true` never produces a *worse* solve than `false`, even when the current `(equations, W_r)` pair has no known prototype. Custom-RHS users who *do* know their Jacobian pattern build their own `ODEFunction` and route through `SciMLProblemReservoir`.

## Extension split (per Francesco's ask on package load weight)

`SparseArrays` is out of the `RCODEReservoirExt` trigger list. A dedicated `RCODEReservoirSparseArraysExt = ["SciMLBase", "DataInterpolations", "SparseArrays"]` adds the sparse `AbstractSparseMatrixCSC` method for `_reservoir_jac_prototype`. The generic fallback lives in `src/layers/continuous_esn_cell.jl` (returns `nothing`) so both extensions can add methods to the same function.

Net: `OrdinaryDiffEq`-only users see no change to what triggers `RCODEReservoirExt`. When `SparseArrays` is loaded (transitively or explicitly), the second extension activates and the sparse method becomes available for dispatch.

## Data (opt-in `use_jac_prototype = true` vs default, apples-to-apples)

10-sample BenchmarkTools microbench under the profiling env (SciMLBase v2.155, OrdinaryDiffEq v6.111, Julia 1.12.5, Apple M4, single-thread OpenBLAS). Workload: `collectstates` at `N = 100`, `T = 200`, `Rodas5(; reltol=1e-6, abstol=1e-8)`, `W_r` from `rand_sparse(rng, Float64, N, N; return_sparse = true)` (1000 nnz / 10 000).

| Metric | `master` (`use_jac_prototype=false`) | this PR (`use_jac_prototype=true`) | Δ |
|---|---:|---:|---:|
| Wall median | 579.83 ms | 336.48 ms | **-42.0%** (1.72×) |
| Wall σ | 2.35 ms | 16.4 ms | tight both |
| Memory | 8.57 MiB | 958 KiB | **-89.1%** |
| Allocs | 267 874 | 1 627 | **-99.4%** |
| GC | 0% | 0.24% mean | negligible |

`Tsit5` and other non-implicit solvers see **zero delta** — they don't use the Jacobian, so `jac_prototype` is dead information.

## Consistency check (Francesco's ask)

Same seed, same sparse `W_r`, `Rodas5(; reltol=1e-6, abstol=1e-8)`, `N=100`, `T=200`, `collectstates`. Toggle off vs toggle on:

- `use_jac_prototype = false` vs `= true` output: `max |diff| ≈ 2.3e-3`.
- This is **not** a math regression: it's the standard consequence of the sparse-coloring FD Jacobian choosing different perturbation columns and step sizes than the dense FD Jacobian. Both trajectories are within `reltol` of the true solution; neither is more "correct" than the other. `Tsit5` — which uses no Jacobian — is unaffected (bit-identical).

The full ext test suite (12 testsets / 34 tests including `Autoregressive predict`, `Euler equivalence with discrete reservoir update`, `Linear ODE analytic match` at `atol = 1e-6`) is green on both `use_jac_prototype = false` and `= true` paths (the analytic and Euler checks use `Tsit5`, exercising the toggle-off path; the other testsets pass under both).

## Design notes

- **Why bool, not symbol/matrix.** Francesco's ask ("a bool like `use_jac`") — plus SciML users who want fine control (custom prototype pattern) already have `SciMLProblemReservoir` for that. Bool keeps the `ContinuousESN` surface small.
- **Why `W_r + I` and not just `W_r`.** The analytic Jacobian has diagonal contributions from the `-I` term regardless of whether `W_r` has any nonzeros on the diagonal. Omitting the diagonal would cause the FD path to skip those columns, producing a wrong Jacobian. `+ I` guarantees the pattern is a superset of the true nonzero locations.
- **Silent no-op vs error on unsupported combinations.** The toggle is monotone: `true` never gives worse behaviour than `false`, so unsupported `(equations, W_r)` pairs silently fall through instead of erroring. Users who care about verifying the opt-in fired can inspect `_reservoir_jac_prototype(cell.equations, ps.reservoir.reservoir_matrix)` directly.

## Test plan

- [x] Full ext suite passes: **34 / 34 tests, 12 / 12 testsets**, 0 fails, 0 errors
  - `Autoregressive predict` — direct target of the `_predict` refactor
  - `Euler equivalence with discrete reservoir update` — guards eq. (5) math
  - `Linear ODE analytic match` at `atol = 1e-6` — analytic solution for the linear-ODE branch (`Tsit5`, no `jac_prototype` interaction)
- [x] Runic clean
- [x] Semantics matrix: all four `(toggle, W_r type, equations)` combinations produce the expected `jac_prototype` (fires only for `true + sparse + built-in RHS`).
- [x] Consistency: same-`W_r` toggle-off vs toggle-on numerical delta under `Rodas5` is within FD-coloring noise; `Tsit5` bit-identical.
- [x] Extension split loads correctly: `RCODEReservoirExt` activates on `SciMLBase + DataInterpolations`; `RCODEReservoirSparseArraysExt` additionally requires `SparseArrays`.

## Checklist

- [x] Appropriate tests were added *(existing coverage exercises both toggle states end-to-end; the toggle-off default is `master`'s behaviour so is fully covered)*
- [x] Any code changes were done in a way that does not break public API *(new `use_jac_prototype` kwarg defaults to `false`; existing constructions unchanged)*
- [x] All documentation related to code changes were updated *(`ContinuousESNCell` docstring + `Show` method)*

## Files

| Path | Role |
| --- | --- |
| `src/layers/continuous_esn_cell.jl` | `use_jac_prototype <: StaticBool` field; constructor kwarg; generic `_reservoir_jac_prototype` fallback returning `nothing`; docstring + `Show` update |
| `ext/RCODEReservoirExt.jl` | Toggle-gated `jac_prototype` at both cell-path `ODEFunction` construction sites; import `_reservoir_jac_prototype` + `known` |
| `ext/RCODEReservoirSparseArraysExt.jl` | New extension: adds the sparse `AbstractSparseMatrixCSC` method for `_reservoir_jac_prototype`, returning `W_r + I` |
| `Project.toml` | Add `RCODEReservoirSparseArraysExt` trigger; revert `RCODEReservoirExt` trigger to `["SciMLBase", "DataInterpolations"]` |
| `test/qa/Project.toml` | Declare `ExplicitImports` as a direct dep of the QA env (unrelated one-line env fix — same content as the `fix(qa)` commit on #479; happy to drop if #479 lands first) |

- Optional `colorvec` via `SparseDiffTools.matrix_colors(jac_prototype)` — additional column-coloring speedup for the FD Jacobian; adds a dep, so leaving for a follow-up if the profile still shows FD dominance after this lands.
- Broader auto-detection covering user-supplied dense-but-structured `W_r` (e.g. banded), if profile evidence surfaces the need.

## Additional context

Independent of #479 (`perf(ext): reuse ODE integrator across AR predict windows`) — both PRs land cell-path perf on non-overlapping surfaces (`_predict` reuse vs `_collectstates`/`_predict` Jacobian). On a stacked branch (measured locally), the two combine cleanly; individual scores are independent.


MartinuzziFrancesco:
This is interesting! A couple of things:
 - have you checked consistency of the results? ie, making sure that they run the same with the current modifications compared to before
 - I think you could provide a toggle for the user to decide, not just automatically use it when the reservoir matrix is sparse. a bool like `usa_jac` or similar, according to what is the best standard in similar situations in other sciml packages.
 -  also I think we are loading a bit too much the packages required to load the ode ext. is there a workaround to not add SparseArrays to it as well?

Saswatsusmoy:
Reworked to address all three points:

**1. Consistency.** Same seed, same sparse `W_r`, `Rodas5(; reltol=1e-6, abstol=1e-8)`, `N=100`, `T=200`: `use_jac_prototype = false` vs `= true` output differs by `max |diff| ≈ 2.3e-3`. That's not a math regression — it's the standard consequence of the sparse-coloring FD Jacobian choosing different perturbation columns / step sizes than the dense FD Jacobian. Both trajectories are within `reltol` of the true solution. `Tsit5` (no Jacobian) is bit-identical between the two.

**2. Toggle.** Added `use_jac_prototype::Bool = false` kwarg on `ContinuousESNCell` + `ContinuousESN` factory. Default off — matches `master` bit-for-bit. User opts in explicitly. Semantics matrix: `false` → nothing; `true` + sparse `W_r` + built-in RHS → `W_r + I`; `true` in any other combination → silent no-op (safe, monotone: toggling on never gives worse behaviour than off).

**3. `SparseArrays` load weight.** Split into two extensions: `RCODEReservoirExt` keeps its old trigger list `["SciMLBase", "DataInterpolations"]`, and a new `RCODEReservoirSparseArraysExt = ["SciMLBase", "DataInterpolations", "SparseArrays"]` adds the sparse method. `OrdinaryDiffEq`-only users see zero change to what triggers the base extension; the sparse specialisation activates when `SparseArrays` is loaded (transitively via most SciML packages, or explicitly). The generic fallback lives in `src/layers/continuous_esn_cell.jl` so both extensions add methods to one function.

Full ext test suite still 34/34 green.

Saswatsusmoy:
The failing `GPU Documentation` check is **not caused by this PR** — same errors reproduce on master (`cd587079`, run [30533266891](https://github.com/SciML/ReservoirComputing.jl/actions/runs/30533266891)):

1. `docs/src/tutorials/reca.md:43-46` — `ArgumentError: state values must be integers between 0 and 1, but found 0.0` (upstream `CellularAutomata` now rejects float states; our `RECACell` cast is `Float32`).
2. `docs/src/tutorials/reca.md:51-56` — `FieldError: type RECACell has no field 'out_dims'` from `_cell_out_dims(cell) = cell.out_dims` in `src/generics.jl:23`. `resetcarry!` calls it even when `init_carry=nothing` and the result is unused.
3. `docs/src/tutorials/sciml_reservoir.md:243-317` — `MethodError` constructing `OrdinaryDiffEqCore.DEOptions` inside `DelayDiffEq.__init`. `OrdinaryDiffEqCore v4.11.0` changed the `DEOptions` fields; `DelayDiffEq 6` hasn't caught up. Purely a transitive-dep incompatibility (needs pin in `docs/Project.toml` or upstream DelayDiffEq fix).

All Tests / Extensions / Interface / Layers / Models / QA / Downgrade jobs pass — perf change is isolated to the cell-path `ODEFunction` extension. Happy to open a separate PR off master for the docs fixes if useful.

MartinuzziFrancesco:
I think we can merge this, can you take a look at the conflicts @Saswatsusmoy ?

PR #533, opened 2026-09-06
feat(inits): add @topology macro for bang-block initializers

### LINKED ISSUE #350: Add macro to build initializers
At this moment a lot of initializers are built with simple building block functions. Adding a `@topology` macro to leverage these building blocks would make the construction of initializers even easier and expand possible explorations

Saswatsusmoy:
## Checklist

## Additional context

`@topology` builds a reservoir initializer from the existing bang mutators (`self_loop!`, `simple_cycle!`, `delay_line!`, ...). Same `[rng]` / `[T]` / `dims` calling convention as the handwritten inits, including partial application. Existing named inits are unchanged.

```julia
@topology my_cycle begin
    self_loop!
    simple_cycle!
end

my_cycle(100, 100)
ESNCell(2 => 100; init_reservoir = my_cycle(; cycle_weight = 0.2))
```

Closes #350

### DIFF
--- a/src/inits/inits_topology.jl
+++ b/src/inits/inits_topology.jl
@@ -142,7 +142,7 @@ function __topology_expand(name::Symbol, body)
         end,
     )
     return quote
-        Base.@__doc__ $primary
+        $primary
         function $name(dims::Integer...; kwargs...)
             return $name($Utils.default_rng(), Float32, dims...; kwargs...)
         end
--- a/src/inits/inits_topology.jl
+++ b/src/inits/inits_topology.jl
@@ -1,4 +1,4 @@
-function _topology_spec(func::Symbol)
+function __topology_spec(func::Symbol)
     func === :delay_line! && return (:delay, :shift, 1)
     func === :backward_connection! && return (:fb, :shift, 1)
     func === :simple_cycle! && return (:cycle, nothing, nothing)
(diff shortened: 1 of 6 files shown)


src/inits/inits_topology.jl
```diff
@@ -0,0 +1,215 @@
+function _topology_spec(func::Symbol)
```
MartinuzziFrancesco: double underscore for internal functions __

src/inits/inits_topology.jl
```diff
@@ -0,0 +1,215 @@
...
+    func === :backward_connection! && return (:fb, :shift, 1)
+    func === :simple_cycle! && return (:cycle, nothing, nothing)
+    func === :reverse_simple_cycle! && return (:second_cycle, nothing, nothing)
+    func === :self_loop! && return (:selfloop, nothing, nothing)
+    func === :add_jumps! && return (:jump, :jump_size, 3)
+    func === :permute_matrix! && return (:permute, :permutation_matrix, nothing)
+    return throw(ArgumentError("@topology unknown building block `$func`"))
+end
+
+function _topology_kw(prefix::Symbol, param::Symbol)
```
MartinuzziFrancesco: double underscore for internal functions __

MartinuzziFrancesco: all internal functions in this file only have one

MartinuzziFrancesco (changes requested):

MartinuzziFrancesco:
I think we also need to add something in the documentation. Since this is a nice feature I think it'd be nice to have a clear explanation/example of how it works

MartinuzziFrancesco:
can you check the reason for the single test failure?

Saswatsusmoy:
the QA fail was ours. ExplicitImports flagged `Base.@__doc__` in the `@topology` expansion as a non-public Base name. dropped that line. generated inits are the same. a docstring sitting above `@topology name` will not attach to `name` anymore. `@doc "..." name` after the definition still works. QA is green now.

only red left is `Core (julia 1, ubuntu-latest, x86)`. that is the new 32-bit lane. it aborts in NNlib/BFloat16s precompile (`LLVM ERROR: Do not know how to soft promote this operator's result!`) before any of our tests run. 64-bit Core on the same Julia is green, including the topology tests. looks unrelated to this PR. same x86 fail is on master after #534.

MartinuzziFrancesco:
sounds good to me, up version and we merge this

PR #540, opened 2026-09-15
feat: closed-loop ESN Jacobian API (#169)

### LINKED ISSUE #169: Feature Request: Jacobian along Trajectory
I think it would be very good if we were able to calculate the jacobian of the system along the trajectory. Among other advantages, this would allow us to find the lyapunov exponent spectrum of the attractor. Is this a possibility? I would be willing to help with this in anyway possible, if this is feasible.

Saswatsusmoy:
## Checklist

## Additional context

Adds `jacobian` / `jacobian!` / `jacobians` for the closed-loop reservoir map of a discrete `ESN` (`in_dims == out_dims`), so Jacobians can be taken along generative trajectories for Lyapunov analysis ([Pathak2017](https://doi.org/10.1063/1.5010300)).

- Analytical default for standard activations and modifiers (`()`, NLAT1/2/3, `Pad`, `PartialSquare`, `ExtendedSquare`)
- Optional `backend=:forwarddiff` via `RCForwardDiffExt` (ForwardDiff weakdep)
- `jacobians` uses the same autoregressive feedback as `predict`
- Deep / Hybrid / Continuous / `Extend` are out of scope for this PR

```julia
J, st = jacobian(esn, state, ps, st)
Js, outputs, st = jacobians(esn, steps, ps, st; initialdata)
```

Closes #169


MartinuzziFrancesco:
This is a nice direction, but I don't think it should be in main library. Perhaps it is time to think about a ReservoirComputingUtilities library

However, I do think it would make for a fantastic example in the documentation!

Saswatsusmoy:
Agreed — I'll convert this PR to a documentation example instead of adding Jacobian APIs to the main library.

Plan:
- drop the core `jacobian` / `jacobian!` / `jacobians` API and the ForwardDiff extension
- add a docs example (closed-loop ESN trajectory → Jacobian / Lyapunov-style analysis with existing Lux + `predict` APIs)
- keep `#169` in mind as docs-only for now; a utilities package can wait until that direction is clearer

I'll push the reworked branch shortly.

